PR #9, opened 2019-11-19
build: add undeclared dependencies to poms

BenWhitehead:
Fixes dependency build failures in #7

### DIFF
--- a/google-cloud-firestore/pom.xml
+++ b/google-cloud-firestore/pom.xml
@@ -57,15 +57,83 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>io.grpc</groupId>
-      <artifactId>grpc-core</artifactId>
-      <version>${grpc.version}</version>
-    </dependency>
-    <dependency>
       <groupId>io.opencensus</groupId>
       <artifactId>opencensus-contrib-grpc-util</artifactId>
       <version>${opencensus.version}</version>
     </dependency>
+    <dependency>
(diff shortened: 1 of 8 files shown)


BenWhitehead:
I've fixed all the warnings from `dependency:analyze` but still getting an error from upper bounds check
```
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for org.codehaus.mojo:animal-sniffer-annotations:1.17 paths to dependency are:
+-com.google.api.grpc:grpc-google-cloud-firestore-admin-v1:1.31.1-SNAPSHOT
  +-io.grpc:grpc-api:1.24.0
    +-org.codehaus.mojo:animal-sniffer-annotations:1.17
and
+-com.google.api.grpc:grpc-google-cloud-firestore-admin-v1:1.31.1-SNAPSHOT
  +-com.google.guava:guava:28.1-android
    +-org.codehaus.mojo:animal-sniffer-annotations:1.18
]
```

I've looked at the animal-sniffer-annotations jars and there is a single annotation in there that is the same between the versions. It is safe to side step it for the time being.

elharo:
Why are we picking up io.grpc:grpc-api:1.24.0? Can this be pushed to 1.25.0?

BenWhitehead:
#5 contains the bump of grpc to 1.25.0 and I think has a break on errorprone. I think I'd prefer to keep it in it's own PR so that it shows up in the changelog whereas this PR should not.

PR #10, opened 2019-11-19
build: update integration project to one with firestore

BenWhitehead:


BenWhitehead:
Can you point me to an example of how to do that?

chingor13:
https://github.com/googleapis/java-bigtable/blob/master/synth.py#L57

BenWhitehead:
Thanks, added the exclude to `synth.py`

PR #53, opened 2019-12-18
build: update annotationProcessorPaths to include auto-service-annotations

BenWhitehead:
With the recent bump to auto-value-annotations 1.7 IntelliJ is not longer
able to build the project due to a NoClassDefFoundError.

Here we manually pull in auto-service-annotations so that it is part of the
processor path because auto-value has it set to provided scope.

This dependency is needed due to the retention change in
https://github.com/google/auto/commit/628df548685b4fc0f2a9af856f97cc2a68da246b
where the RetentionPolicy changed from SOURCE to CLASS.

Due to the RetentionPolicy change to CLASS we must have the
annotations available on the processor path otherwise the following
will error will be thrown. (This is a particular problem with the
annotation processor configuration in IntelliJ)

Error:java: java.lang.NoClassDefFoundError: com/google/auto/service/AutoService
com.google.auto.service.AutoService

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -327,20 +327,6 @@
         </plugin>
       </plugins>
     </pluginManagement>
-    <plugins>
-      <plugin>
-        <artifactId>maven-compiler-plugin</artifactId>
-        <configuration>
-          <annotationProcessorPaths>
-            <path>
-              <groupId>com.google.auto.value</groupId>
-              <artifactId>auto-value</artifactId>
-              <version>${auto-value.version}</version>
-            </path>
-          </annotationProcessorPaths>
-        </configuration>
-      </plugin>
-    </plugins>
   </build>
 
   <modules>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -364,6 +350,29 @@
...
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+              <annotationProcessorPaths>
+                <path>
+                  <groupId>com.google.auto.value</groupId>
+                  <artifactId>auto-value</artifactId>
+                  <version>${auto-value.version}</version>
+                </path>
+                <!--
+                There is currently no available version of auto-service-annotations
```
elharo: The Java 7 CI failed. Perhaps we should ask @netdpb to push one compiled for Java 7?

pom.xml
```diff
@@ -364,6 +350,29 @@
         <auto-value-annotation.version>1.6.6</auto-value-annotation.version>
```
elharo: Isn't this up to 1.7 now?

pom.xml
```diff
@@ -374,7 +383,45 @@
       <properties>
         <auto-value.version>1.6.6</auto-value.version>
```
elharo: 1.7?

pom.xml
BenWhitehead: Good catch, I kept bouncing back and forth I've changed to 1.7.

pom.xml
BenWhitehead: The java 7 build failed because of a timeout from maven central.

I don't know if we specifically need a release for Java 7 as we perform release builds with Java 8 and a 1.7 target.

netdpb:
IntelliJ gives no further information about that error, like a stacktrace? `@AutoService` has class retention; it shouldn't have to be on the classpath to load the AutoValue processor.

BenWhitehead:
Correct @netdpb This is all that was printed by IntelliJ, even with `-verbose` added to the compiler args.

PR #103, opened 2020-02-11
feat: add new Firestore.runAsyncTransaction

### LINKED ISSUE #43: Firestore.runTransaction should support/recognize asynchronous updateFunction result
Thanks for creating and managing Firestore. I have one suggestion for improvement though.

Currently I consider Firestore admin API to be just partially asynchronous. Whenever there is a call which returns ApiFuture, I consider it OK, although far from ideal comparing to RxJava/reactor Single/Observable/Mono/Flux. I can do conversion from ApiFuture to Mono quite easily by myself.

### Problem
The Firestore.runTransaction() has a problem: result of the runTransaction is OK (it's ApiFuture). The transaction update function, however, must provide a result synchronously:

```
  public interface Function<T> {

    T updateCallback(Transaction transaction) throws Exception;
  }

```
When I track down to the point where the function is executed, I observe:

```
          private SettableApiFuture<T> invokeUserCallback() {
            // Execute the user callback on the provided executor.
            final SettableApiFuture<T> callbackResult = SettableApiFuture.create();
            userCallbackExecutor.execute(
                new Runnable() {
                  @Override
                  public void run() {
                    try {
                      callbackResult.set(transactionCallback.updateCallback(transaction));
                    } catch (Throwable t) {
                      callbackResult.setException(t);
                    }
                  }
                });
            return callbackResult;
          }

```
This implies, that that whole body of the update function is executed in a blocking way, thus wasting precious resources.

### Resolution
My suggestion for improvement is instead of calling callbackResult.set(transactionCallback.updateCallback(transaction)), it would be better to distinguish asynchronous result and do the invocation specifically:

```
            userCallbackExecutor.execute(
                new Runnable() {
                  @Override
                  public void run() {

                    try {
                      T callbackValue = transactionCallback.updateCallback(transaction);
                      if (callbackValue instanceof ApiFuture) {
                        ApiFutures.addCallback((ApiFuture<T>) callbackValue, new ApiFutureCallback<T>() {
                          @Override
                          public void onFailure(Throwable throwable) {
                            callbackResult.setException(throwable);
                          }

                          @Override
                          public void onSuccess(T t) {
                            callbackResult.set(t);
                          }
                        });
                      } else {
                        callbackResult.set(callbackValue);
                      }
                    } catch (Throwable t) {
                      callbackResult.setException(t);
                    }
                  }
                });

```
With this implementation, I could implement body of the update function in a fully reactive way and convert the result to ApiFuture (SettableApiFuture).

### Alternatives
A cleaner solution - describing the contract more clearly - would be to have separate interface for asynchronous update function:

```
  public interface AsyncFunction<T> {

    ApiFuture<T> updateCallbackAsync(Transaction transaction) throws Exception;
  }

```
And then specific method for executing AsyncFunction instead of Function. Actually anything which somehow allows me to do asynchronous transaction execution would offer me the required flexibility/efficiency.

Another alternative would be to provide fully reactive API (returning Mono/Single), but I understand that you do not want dependency on particular vendor (RxJava, Spring reactor). Maybe, however, you could consider switching to reactive streams Publisher, which is becoming the common shared API/standard for reactive programming in Java.

kvakos:
Fixes #43 (it's a good idea to open an issue first for context and/or discussion)

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
@@ -100,6 +100,43 @@ public interface Firestore extends Service<FirestoreOptions>, AutoCloseable {
       @Nonnull TransactionOptions transactionOptions);
 
   /**
+   * Executes the given updateFunction and then attempts to commit the changes applied within the
+   * transaction. If any document read within the transaction has changed, the updateFunction will
+   * be retried. If it fails to commit after 5 attempts, the transaction will fail. <br>
+   * <br>
+   * Running a transaction places locks all consumed documents. To unblock other clients, the
+   * Firestore backend automatically releases all locks after 60 seconds of inactivity and fails all
+   * transactions that last longer than 270 seconds (see <a
+   * href="https://firebase.google.com/docs/firestore/quotas#writes_and_transactions">Firestore
+   * Quotas</a>).
+   *
+   * @param updateFunction The function to execute within the transaction context.
+   * @return An ApiFuture that will be resolved with the result from updateFunction.
+   */
+  @Nonnull
+  <T> ApiFuture<T> runAsyncTransaction(@Nonnull final Transaction.AsyncFunction<T> updateFunction);
+
+  /**
+   * Executes the given updateFunction and then attempts to commit the changes applied within the
+   * transaction. If any document read within the transaction has changed, the updateFunction will
+   * be retried. If it fails to commit after the maxmimum number of attemps specified in
+   * transactionOptions, the transaction will fail. <br>
+   * <br>
+   * Running a transaction places locks all consumed documents. To unblock other clients, the
+   * Firestore backend automatically releases all locks after 60 seconds of inactivity and fails all
+   * transactions that last longer than 270 seconds (see <a
+   * href="https://firebase.google.com/docs/firestore/quotas#writes_and_transactions">Firestore
+   * Quotas</a>).
+   *
+   * @param updateFunction The function to execute within the transaction context.
+   * @return An ApiFuture that will be resolved with the result from updateFunction.
+   */
+  @Nonnull
+  <T> ApiFuture<T> runAsyncTransaction(
+      @Nonnull final Transaction.AsyncFunction<T> updateFunction,
+      @Nonnull TransactionOptions transactionOptions);
+
+  /**
    * Retrieves multiple documents from Firestore.
    *
    * @param documentReferences List of Document References to fetch.
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -299,13 +299,30 @@ class FirestoreImpl implements Firestore {
       @Nonnull final Transaction.Function<T> updateFunction,
       @Nonnull TransactionOptions transactionOptions) {
     SettableApiFuture<T> resultFuture = SettableApiFuture.create();
+    runTransaction(new TransactionAsyncAdapter<>(updateFunction), resultFuture, transactionOptions);
+    return resultFuture;
+  }
+
+  @Nonnull
+  @Override
+  public <T> ApiFuture<T> runAsyncTransaction(
+      @Nonnull final Transaction.AsyncFunction<T> updateFunction) {
+    return runAsyncTransaction(updateFunction, TransactionOptions.create());
+  }
+
+  @Nonnull
+  @Override
+  public <T> ApiFuture<T> runAsyncTransaction(
+      @Nonnull final Transaction.AsyncFunction<T> updateFunction,
+      @Nonnull TransactionOptions transactionOptions) {
+    SettableApiFuture<T> resultFuture = SettableApiFuture.create();
     runTransaction(updateFunction, resultFuture, transactionOptions);
     return resultFuture;
   }
 
   /** Transaction functions that returns its result in the provided SettableFuture. */
   private <T> void runTransaction(
-      final Transaction.Function<T> transactionCallback,
+      final Transaction.AsyncFunction<T> transactionCallback,
       final SettableApiFuture<T> resultFuture,
       final TransactionOptions options) {
     // span is intentionally not ended here. It will be ended by runTransactionAttempt on success
@@ -317,7 +334,7 @@ class FirestoreImpl implements Firestore {
   }
 
   private <T> void runTransactionAttempt(
-      final Transaction.Function<T> transactionCallback,
+      final Transaction.AsyncFunction<T> transactionCallback,
       final SettableApiFuture<T> resultFuture,
       final TransactionOptions options,
       final Span span) {
@@ -384,7 +401,21 @@ class FirestoreImpl implements Firestore {
                   @Override
                   public void run() {
                     try {
-                      callbackResult.set(transactionCallback.updateCallback(transaction));
+                      ApiFuture<T> updateCallback = transactionCallback.updateCallback(transaction);
+                      ApiFutures.addCallback(
+                          updateCallback,
+                          new ApiFutureCallback<T>() {
+                            @Override
+                            public void onFailure(Throwable t) {
+                              callbackResult.setException(t);
+                            }
+
+                            @Override
+                            public void onSuccess(T result) {
+                              callbackResult.set(result);
+                            }
+                          },
+                          MoreExecutors.directExecutor());
                     } catch (Throwable t) {
                       callbackResult.setException(t);
                     }
@@ -493,5 +524,24 @@ class FirestoreImpl implements Firestore {
   public void close() throws Exception {
     firestoreClient.close();
     closed = true;
+  }
+
+  private static class TransactionAsyncAdapter<T> implements Transaction.AsyncFunction<T> {
+    private final Transaction.Function<T> syncFunction;
+
+    public TransactionAsyncAdapter(Transaction.Function<T> syncFunction) {
+      this.syncFunction = syncFunction;
+    }
+
+    @Override
+    public ApiFuture<T> updateCallback(Transaction transaction) {
+      SettableApiFuture<T> callbackResult = SettableApiFuture.create();
+      try {
+        callbackResult.set(syncFunction.updateCallback(transaction));
+      } catch (Throwable e) {
+        callbackResult.setException(e);
+      }
+      return callbackResult;
+    }
   }
 }
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
@@ -42,13 +42,23 @@ public final class Transaction extends UpdateBuilder<Transaction> {
       "Firestore transactions require all reads to be executed before all writes";
 
   /**
-   * User callback that takes a Firestore Transaction
+   * User callback that takes a Firestore Transaction.
    *
    * @param <T> The result type of the user callback.
    */
   public interface Function<T> {
 
     T updateCallback(Transaction transaction) throws Exception;
+  }
+
+  /**
+   * User callback that takes a Firestore Async Transaction.
+   *
+   * @param <T> The result type of the user async callback.
+   */
+  public interface AsyncFunction<T> {
+
+    ApiFuture<T> updateCallback(Transaction transaction);
   }
 
   private final ByteString previousTransactionId;
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
@@ -34,6 +34,7 @@ import static com.google.cloud.firestore.LocalFirestoreHelper.rollbackResponse;
 import static com.google.cloud.firestore.LocalFirestoreHelper.set;
 import static com.google.cloud.firestore.LocalFirestoreHelper.update;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.doAnswer;
@@ -133,6 +134,33 @@ public class TransactionTest {
   }
 
   @Test
+  public void returnsValueAsync() throws Exception {
+    doReturn(beginResponse())
+        .doReturn(commitResponse(0, 0))
+        .when(firestoreMock)
+        .sendRequest(requestCapture.capture(), Matchers.<UnaryCallable<Message, Message>>any());
+
+    ApiFuture<String> transaction =
+        firestoreMock.runAsyncTransaction(
+            new Transaction.AsyncFunction<String>() {
+              @Override
+              public ApiFuture<String> updateCallback(Transaction transaction) {
+                Assert.assertEquals("user_provided", Thread.currentThread().getName());
+                return ApiFutures.immediateFuture("foo");
+              }
+            },
+            options);
+
+    assertEquals("foo", transaction.get());
+
+    List<Message> requests = requestCapture.getAllValues();
+    assertEquals(2, requests.size());
+
+    assertEquals(begin(), requests.get(0));
+    assertEquals(commit(TRANSACTION_ID), requests.get(1));
+  }
+
+  @Test
   public void canReturnNull() throws Exception {
     doReturn(beginResponse())
         .doReturn(ApiFutures.immediateFailedFuture(new Exception()))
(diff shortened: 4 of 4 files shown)


kvakos:
@googlebot I signed it!

kvakos:
@googlebot I signed it!

kvakos:
@googlebot I signed it!

kvakos:
@googlebot I signed it!

kvakos:
I cannot pass Binary Compatibility test. any suggestions?

BenWhitehead:
Binary compatibility has broken due to the new interface methods, which would cause any existing implementers of the interface to no longer satisfy the interfaces requirements.
```
[ERROR] 7012: com.google.cloud.firestore.Firestore: Method 'public com.google.api.core.ApiFuture runTransactionAsync(com.google.cloud.firestore.Transaction$AsyncFunction)' has been added to an interface
[ERROR] 7012: com.google.cloud.firestore.Firestore: Method 'public com.google.api.core.ApiFuture runTransactionAsync(com.google.cloud.firestore.Transaction$AsyncFunction, com.google.cloud.firestore.TransactionOptions)' has been added to an interface
```

@chingor13 What is our general policy around adding to veneer interfaces? I know in generated protos we've been a bit more permissive.

@schmidt-sebastian Can you please review this contribution? I believe you're more familiar with the transaction handling logic in the code than I am.

BenWhitehead:
(@schmidt-sebastian I can't assign this PR to you right now, have to track down some permissions)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
```diff
@@ -51,6 +51,16 @@
     T updateCallback(Transaction transaction) throws Exception;
   }
 
+  /**
+   * User callback that takes a Firestore Transaction
```
schmidt-sebastian: Please update to mention the async nature of this callback.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -260,6 +373,55 @@ public String updateCallback(Transaction transaction) {
     assertEquals(commit(ByteString.copyFromUtf8("foo5")), requests.get(9));
   }
 
+  @Test
+  public void limitsRetriesWithFailureAsync() throws Exception {
```
schmidt-sebastian: This test doesn't seem to add test coverage over the existing `limitsRetriesWithFailure()` test. If you agree, we could probably remove it. Same applies to `limitsRetriesWithSuccessAsync()`.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -384,6 +630,46 @@ public void getMultipleDocuments() throws Exception {
     assertEquals(commit(TRANSACTION_ID), requests.get(2));
   }
 
+  @Test
+  public void getMultipleDocumentsAsync() throws Exception {
```
schmidt-sebastian: The above comment also applies here. The code path you are testing is already covered by `getDocumentAsync()`. This also applies to other test functions added below.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -85,6 +85,31 @@
...
+  /**
+   * Executes the given updateFunction and then attempts to commit the changes applied within the
+   * transaction. If any document read within the transaction has changed, the updateFunction will
+   * be retried. If it fails to commit after 5 attempts, the transaction will fail.
+   *
+   * @param updateFunction The function to execute within the transaction context.
+   * @return An ApiFuture that will be resolved with the result from updateFunction.
+   */
+  @Nonnull
+  <T> ApiFuture<T> runTransactionAsync(@Nonnull final Transaction.AsyncFunction<T> updateFunction);
```
schmidt-sebastian: We use the "async" suffix in other parts of the API to indicate that the function returns an ApiFuture. "Async" is usually coupled with a blocking API that lacks the suffix. I think what you are doing here is slightly different - `runTransaction()` is also async, but in a slightly different way.

What do you think of renaming these new methods to `runAsyncTransaction`?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -85,6 +85,31 @@
   @Nonnull
   <T> ApiFuture<T> runTransaction(@Nonnull final Transaction.Function<T> updateFunction);
 
+  /**
+   * Executes the given updateFunction and then attempts to commit the changes applied within the
+   * transaction. If any document read within the transaction has changed, the updateFunction will
+   * be retried. If it fails to commit after 5 attempts, the transaction will fail.
+   *
+   * @param updateFunction The function to execute within the transaction context.
```
schmidt-sebastian: One potential downside of your proposal is that it makes it now much easier to invoke transactions that take a long time to execute (as the user may now be inclined to kick off other tasks as well). Can you add something to the comment to explain these downsides? E.g. something like: "Running a transaction places locks all all consumed documents. To unblock other clients, the Firestore backend automatically releases all locks after 60 seconds of inactivity and fails all transactions that last longer than 270 seconds (see https://firebase.google.com/docs/firestore/quotas#writes_and_transactions)"

schmidt-sebastian:
> @chingor13 What is our general policy around adding to veneer interfaces? I know in generated protos we've been a bit more permissive.

I would not consider implementing a custom FirebaseFirestore a supported use case. While it is true that it might break some of our users, this limitation will make it very hard for us to add additional functionality.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
kvakos: comment added to the javadoc

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
kvakos: method renamed

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
kvakos: test removed

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
kvakos: tests removed

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
kvakos: check it, please

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -85,6 +85,43 @@
...
+   * Quotas</a>).
+   *
+   * @param updateFunction The function to execute within the transaction context.
+   * @return An ApiFuture that will be resolved with the result from updateFunction.
+   */
+  @Nonnull
+  <T> ApiFuture<T> runAsyncTransaction(
+      @Nonnull final Transaction.AsyncFunction<T> updateFunction,
+      @Nonnull TransactionOptions transactionOptions);
+
```
schmidt-sebastian: Please move both these methods below the overload for `runTransaction`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -384,7 +401,21 @@ public void onSuccess(List<WriteResult> writeResults) {
                   @Override
                   public void run() {
                     try {
-                      callbackResult.set(transactionCallback.updateCallback(transaction));
+                      ApiFuture<T> updateCallback = transactionCallback.updateCallback(transaction);
+                      ApiFutures.addCallback(
```
schmidt-sebastian: If I am reading the documentation of `addCallback` correctly in only runs the `ApiFutureCallback` on the provided executor. We actually want to run `transactionCallback` on the user executor. Do you mind fixing that and/or convincing me that I am mistaken?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
```diff
@@ -51,6 +51,16 @@
     T updateCallback(Transaction transaction) throws Exception;
   }
 
+  /**
+   * User callback that takes a Firestore Async Transaction
```
schmidt-sebastian: "Async user callback that takes a Firestore transaction."

Can you also add a period to line 45?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -33,14 +33,13 @@
 import static com.google.cloud.firestore.LocalFirestoreHelper.rollbackResponse;
 import static com.google.cloud.firestore.LocalFirestoreHelper.set;
 import static com.google.cloud.firestore.LocalFirestoreHelper.update;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
schmidt-sebastian: Please no glob imports.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -213,6 +298,34 @@ public String updateCallback(Transaction transaction) {
     assertEquals(1, requests.size());
   }
 
+  @Test
+  public void noRollbackOnBeginFailureAsync() throws Exception {
```
schmidt-sebastian: You didn't change this code path, so I think we can remove this test.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -428,6 +575,49 @@ public void getMultipleDocumentsWithFieldMask() throws Exception {
     assertEquals(commit(TRANSACTION_ID), requests.get(2));
   }
 
+  @Test
+  public void getMultipleDocumentsWithFieldMaskAsync() throws Exception {
```
schmidt-sebastian: Your PRs changes are already tested by `getDocumentAsync()`. Since there is  no specific handling of GetAll, I would suggest to remove this test.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -463,6 +653,40 @@ public QuerySnapshot updateCallback(Transaction transaction)
     assertEquals(commit(TRANSACTION_ID), requests.get(2));
   }
 
+  @Test
+  public void getQueryAsync() throws Exception {
```
schmidt-sebastian: Same comment as above.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -497,6 +721,42 @@ public String updateCallback(Transaction transaction) {
     assertEquals(commit(TRANSACTION_ID, writes.toArray(new Write[] {})), requests.get(1));
   }
 
+  @Test
+  public void updateDocumentAsync() throws Exception {
```
schmidt-sebastian: Same comment as above.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -532,6 +792,43 @@ public String updateCallback(Transaction transaction) {
     assertEquals(commit(TRANSACTION_ID, writes.toArray(new Write[] {})), requests.get(1));
   }
 
+  @Test
+  public void setDocumentAsync() throws Exception {
```
schmidt-sebastian: Same comment as above.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -567,6 +864,43 @@ public String updateCallback(Transaction transaction) {
     assertEquals(commit(TRANSACTION_ID, writes.toArray(new Write[] {})), requests.get(1));
   }
 
+  @Test
+  public void createDocumentAsync() throws Exception {
```
schmidt-sebastian: Same comment as above.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -604,4 +938,44 @@ public String updateCallback(Transaction transaction) {
     assertEquals(begin(), requests.get(0));
     assertEquals(commit(TRANSACTION_ID, writes.toArray(new Write[] {})), requests.get(1));
   }
+
+  @Test
+  public void deleteDocumentAsync() throws Exception {
```
schmidt-sebastian: Same comment as above.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
kvakos: I duplicate all tests in TransactionTest.java that somehow play with runTransaction and implemented them with runAsyncTransaction. Old sync/blocking methods as it is now implemented also are testing async implementation. I can delete all of *async tests. Can I delete all *async  tests?

kvakos: Almost all *async tests were deleted except those with other mocked setup

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
kvakos: You are right.

kvakos:
Can I resolve done conversation?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -132,6 +134,35 @@ public String updateCallback(Transaction transaction) {
...
+
+    ApiFuture<String> transaction =
+        firestoreMock.runAsyncTransaction(
+            new Transaction.AsyncFunction<String>() {
+              @Override
+              public ApiFuture<String> updateCallback(Transaction transaction) {
+                Assert.assertEquals("user_provided", Thread.currentThread().getName());
+                SettableApiFuture<String> apiFuture = SettableApiFuture.create();
+                apiFuture.set("foo");
+                return apiFuture;
```
schmidt-sebastian: Nit: Replace with https://github.com/googleapis/api-common-java/blob/master/src/main/java/com/google/api/core/ApiFutures.java#L107 (here and below)

kvakos: replaced both places with ApiFutures#immediateFuture()

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
kvakos: moved

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
kvakos: period was added to both

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
kvakos: global import replaced

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
kvakos: removed

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
kvakos: removed

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
kvakos: removed

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
kvakos: removed

google-cloud-firestore/src/test/java/com/google/cloud/firestore/TransactionTest.java
```diff
@@ -185,6 +236,39 @@ public String updateCallback(Transaction transaction) throws Exception {
...
+        .sendRequest(requestCapture.capture(), Matchers.<UnaryCallable<Message, Message>>any());
+
+    ApiFuture<String> transaction =
+        firestoreMock.runAsyncTransaction(
+            new Transaction.AsyncFunction<String>() {
+              @Override
+              public ApiFuture<String> updateCallback(Transaction transaction) {
+                SettableApiFuture<String> apiFuture = SettableApiFuture.create();
+                apiFuture.setException(new Exception("Expected exception"));
+                return apiFuture;
```
schmidt-sebastian: You can use an "immediateFailedFuture" here if you want to do another edit :)

kvakos:
To @schmidt-sebastian: I appreciate very much your help. Thanks a lot.

BenWhitehead:
@kvakos Can you run `mvn com.coveo:fmt-maven-plugin:format` to fixup the source files for the `Code Format` check?

kvakos:
Code format applied

BenWhitehead:
Thanks @kvakos I'll follow up and get info about how the new interface methods will impact versioning.

BenWhitehead:
@kvakos We've figured out what needs to be done as far as versioning is concerned and we should be able to merge this soon (most likely as is, at worst after a rebase).

BenWhitehead:
I'll fix the binary compatibility build shortly in another PR

PR #107, opened 2020-02-13
fix: deflake-ify ITSystemTest#queryWatch

BenWhitehead:
queryWatch has been flaky because several things are happening in the
single event stream. This fix splits up the different scenarios being
tested into their own respective tests and updates the assertions to be
more specific to the scenario as well as handling the semantics of the
backend better.

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
@@ -385,6 +385,8 @@ public final class ITQueryWatchTest {
 
   private static final class EventsCountDownLatch {
     private final CountDownLatch initialEventsCountDownLatch;
+    private final int initialEventCount;
+    private final EnumMap<DocumentChange.Type, Integer> eventsCounts;
     private final EnumMap<DocumentChange.Type, CountDownLatch> eventsCountDownLatches;
 
     EventsCountDownLatch(
@@ -393,6 +395,11 @@ public final class ITQueryWatchTest {
         int modifiedInitialCount,
         int removedInitialCount) {
       initialEventsCountDownLatch = new CountDownLatch(initialEventCount);
+      this.initialEventCount = initialEventCount;
+      eventsCounts = new EnumMap<>(DocumentChange.Type.class);
+      eventsCounts.put(DocumentChange.Type.ADDED, addedInitialCount);
+      eventsCounts.put(DocumentChange.Type.MODIFIED, modifiedInitialCount);
+      eventsCounts.put(DocumentChange.Type.REMOVED, removedInitialCount);
       eventsCountDownLatches = new EnumMap<>(DocumentChange.Type.class);
       eventsCountDownLatches.put(DocumentChange.Type.ADDED, new CountDownLatch(addedInitialCount));
       eventsCountDownLatches.put(
@@ -410,11 +417,12 @@ public final class ITQueryWatchTest {
     }
 
     void awaitInitialEvents() throws InterruptedException {
-      initialEventsCountDownLatch.await(5, TimeUnit.SECONDS);
+      initialEventsCountDownLatch.await(5 * initialEventCount, TimeUnit.SECONDS);
     }
 
     void await(DocumentChange.Type type) throws InterruptedException {
-      eventsCountDownLatches.get(type).await(5, TimeUnit.SECONDS);
+      int count = eventsCounts.get(type);
+      eventsCountDownLatches.get(type).await(5 * count, TimeUnit.SECONDS);
     }
   }
 
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
@@ -87,9 +87,13 @@ public final class ITQueryWatchTest {
     firestore.close();
   }
 
-  /*
-  Attach a listener to a query with empty results.
-  Verify the listener receives an empty event.
+  /**
+   *
+   *
+   * <ol>
+   *   <li>Attach a listener to a query with empty results.
+   *   <li>Verify the listener receives an empty event.
+   * </ol>
    */
   @Test
   public void emptyResults() throws InterruptedException {
@@ -112,9 +116,13 @@ public final class ITQueryWatchTest {
     listenerAssertions.removedIdsIsAnyOf(emptySet());
   }
 
-  /*
-  Attach a listener to a query with non-empty results.
-  Verify the listener receives an event including the expected document.
+  /**
+   *
+   *
+   * <ol>
+   *   <li>Attach a listener to a query with non-empty results.
+   *   <li>Verify the listener receives an event including the expected document.
+   * </ol>
    */
   @Test
   public void nonEmptyResults() throws InterruptedException, TimeoutException, ExecutionException {
@@ -139,11 +147,15 @@ public final class ITQueryWatchTest {
     listenerAssertions.removedIdsIsAnyOf(emptySet());
   }
 
-  /*
-  Attach a listener to a query with empty results.
-  Create a new document that matches the query.
-  Verify newly created document results in an ADDED event.
-  */
+  /**
+   *
+   *
+   * <ol>
+   *   <li>Attach a listener to a query with empty results.
+   *   <li>Create a new document that matches the query.
+   *   <li>Verify newly created document results in an ADDED event.
+   * </ol>
+   */
   @Test
   public void emptyResults_newDocument_ADDED()
       throws InterruptedException, TimeoutException, ExecutionException {
@@ -168,11 +180,15 @@ public final class ITQueryWatchTest {
     listenerAssertions.removedIdsIsAnyOf(emptySet());
   }
 
-  /*
-  Attach a listener to a query with empty results.
-  Modify an existing document so that it matches the query.
-  Verify newly created document results in an ADDED event.
-  */
+  /**
+   *
+   *
+   * <ol>
+   *   <li>Attach a listener to a query with empty results.
+   *   <li>Modify an existing document so that it matches the query.
+   *   <li>Verify newly created document results in an ADDED event.
+   * </ol>
+   */
   @Test
   public void emptyResults_modifiedDocument_ADDED()
       throws InterruptedException, TimeoutException, ExecutionException {
@@ -206,11 +222,15 @@ public final class ITQueryWatchTest {
     assertThat(doc.get("baz")).isEqualTo("baz");
   }
 
-  /*
-  Attach a listener to a query with non-empty results.
-  Modify an existing document that is part of the results.
-  Verify modified document results in a MODIFIED event.
-  */
+  /**
+   *
+   *
+   * <ol>
+   *   <li>Attach a listener to a query with non-empty results.
+   *   <li>Modify an existing document that is part of the results.
+   *   <li>Verify modified document results in a MODIFIED event.
+   * </ol>
+   */
   @Test
   public void nonEmptyResults_modifiedDocument_MODIFIED()
       throws InterruptedException, TimeoutException, ExecutionException {
@@ -246,11 +266,15 @@ public final class ITQueryWatchTest {
     assertThat(doc.get("baz")).isEqualTo("baz");
   }
 
-  /*
-  Attach a listener to a query with non-empty results.
-  Delete an existing document that is part of the results.
-  Verify deleted document results in a REMOVED event.
-  */
+  /**
+   *
+   *
+   * <ol>
+   *   <li>Attach a listener to a query with non-empty results.
+   *   <li>Delete an existing document that is part of the results.
+   *   <li>Verify deleted document results in a REMOVED event.
+   * </ol>
+   */
   @Test
   public void nonEmptyResults_deletedDocument_REMOVED()
       throws InterruptedException, TimeoutException, ExecutionException {
@@ -285,11 +309,15 @@ public final class ITQueryWatchTest {
     assertThat(doc.get("foo")).isEqualTo("bar");
   }
 
-  /*
-  Attach a listener to a query with non-empty results.
-  Modify an existing document that is part of the results to no longer match the query.
-  Verify modified document results in a REMOVED event.
-  */
+  /**
+   *
+   *
+   * <ol>
+   *   <li>Attach a listener to a query with non-empty results.
+   *   <li>Modify an existing document that is part of the results to no longer match the query.
+   *   <li>Verify modified document results in a REMOVED event.
+   * </ol>
+   */
   @Test
   public void nonEmptyResults_modifiedDocument_REMOVED()
       throws InterruptedException, TimeoutException, ExecutionException {
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
@@ -102,9 +102,9 @@ public final class ITQueryWatchTest {
     }
 
     assertThat(listener.receivedEvents).hasSize(1);
-    ListenerEvent e = listener.receivedEvents.get(0);
-    assertThat(e.error).isNull();
-    QuerySnapshot snap = e.value;
+    ListenerEvent event = listener.receivedEvents.get(0);
+    assertThat(event.error).isNull();
+    QuerySnapshot snap = event.value;
     assertThat(snap).isNotNull();
     assertThat(snap).isEmpty();
     assertThat(snap.getDocumentChanges()).isEmpty();
@@ -131,9 +131,9 @@ public final class ITQueryWatchTest {
     }
 
     assertThat(receivedEvents).hasSize(1);
-    ListenerEvent e = receivedEvents.get(0);
-    assertThat(e.error).isNull();
-    QuerySnapshot snap = e.value;
+    ListenerEvent event = receivedEvents.get(0);
+    assertThat(event.error).isNull();
+    QuerySnapshot snap = event.value;
     assertThat(snap).isNotNull();
     assertThat(snap).hasSize(1);
   }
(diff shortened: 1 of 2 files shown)


kolea2:
LGTM

BenWhitehead:
@schmidt-sebastian Can you get a quick review from you on this PR?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+    randomColl = firestore.collection(collPath);
+  }
+
+  @AfterClass
+  public static void afterClass() throws Exception {
+    firestore.close();
+  }
+
+  /*
+  Start a listener on a query with an empty result set
```
schmidt-sebastian: Please fix (here and everywhere).

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+    ListenerEvent e = receivedEvents.get(0);
+    assertThat(e.error).isNull();
+    QuerySnapshot snap = e.value;
+    assertThat(snap).isNotNull();
+    assertThat(snap).hasSize(1);
+  }
+
+  /*
+  Starting from a query with empty results creating a new document
+  that matches the query should result in an ADDED event
```
schmidt-sebastian: As you fix the comments, do you mind splitting this into two sentences?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+   */
+  @Test
+  public void emptyResults_newDocument_ADDED()
```
schmidt-sebastian: This name could use a sprinkle of the Java Style Guide :)

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+
+  /*
+  Start a listener on a query with a non-empty result set
+   */
+  @Test
+  public void nonEmptyResults() throws InterruptedException, TimeoutException, ExecutionException {
+    // create a document in our collection that will match the query
+    randomColl.document("doc").set(map("foo", "bar")).get(5, TimeUnit.SECONDS);
+
+    // run our test
```
schmidt-sebastian: Nit: I think the "run our test" comments can be removed.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+    try {
+      randomColl.document("doc").set(map("foo", "bar")).get(5, TimeUnit.SECONDS);
+      listener.eventsCDL.await(DocumentChange.Type.ADDED);
+    } finally {
+      registration.remove();
+    }
+
+    ListenerAssertions la = listener.assertions();
+    la.noError();
+    la.eventCountIsAnyOf(Range.closed(1, 2));
```
schmidt-sebastian: If you wait for the initial empty event, the number of events is deterministic.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+  public void emptyResults_newDocument_ADDED()
+      throws InterruptedException, TimeoutException, ExecutionException {
+
+    // run our test
+    final Query query = randomColl.whereEqualTo("foo", "bar");
+    QuerySnapshotEventListener listener = new QuerySnapshotEventListener(1, 1, 0, 0);
+    ListenerRegistration registration = query.addSnapshotListener(listener);
+
+    try {
+      randomColl.document("doc").set(map("foo", "bar")).get(5, TimeUnit.SECONDS);
```
schmidt-sebastian: You don't need to block on this - the write doesn't need to be finished for the listener to fire (the two streams are independent).

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+    la.noError();
+    la.eventCountIsAnyOf(Range.closed(1, 2));
+    la.addedIdsIsAnyOf(emptySet(), newHashSet("doc"));
+    la.modifiedIdsIsAnyOf(emptySet());
+    la.removedIdsIsAnyOf(emptySet());
+  }
+
+  /*
+  Starting from a query with empty results, modifying an existing document to
+  match the query should result in an ADDED event
```
schmidt-sebastian: Please split into two sentences.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+   */
+  @Test
+  public void emptyResults_modifiedDocument_ADDED()
```
schmidt-sebastian: Same comment regarding the style guide.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+    // create our "existing non-matching document"
+    randomColl.document("doc").set(map("baz", "baz")).get(5, TimeUnit.SECONDS);
+
+    // run our test
+    final Query query = randomColl.whereEqualTo("foo", "bar");
+    QuerySnapshotEventListener listener = new QuerySnapshotEventListener(1, 1, 0, 0);
+    List<ListenerEvent> receivedEvents = listener.receivedEvents;
+    ListenerRegistration registration = query.addSnapshotListener(listener);
+
+    try {
```
schmidt-sebastian: This test would be easier to follow if you waited for the initial snapshot. Do you mind updating all tests?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+      listener.eventsCDL.await();
+    } finally {
+      registration.remove();
+    }
+
+    assertThat(listener.receivedEvents).hasSize(1);
+    ListenerEvent e = listener.receivedEvents.get(0);
```
schmidt-sebastian: Nit: Java generally doesn't use abbreviations like this (here and below).

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+    @Nullable private final QuerySnapshot value;
+    @Nullable private final FirestoreException error;
+
+    ListenerEvent(@Nullable QuerySnapshot value, @Nullable FirestoreException error) {
+      this.value = value;
+      this.error = error;
+    }
+  }
+
+  private static final class EventsCDL {
```
schmidt-sebastian: Please don't use acronyms. We need to feed the stereotype that Java is much more verbose that C++.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,484 @@
...
+    private final CountDownLatch cdl;
+    private final EnumMap<DocumentChange.Type, CountDownLatch> cdls;
+
+    EventsCDL(int nonChangeInitial, int addedInitial, int modifiedInitial, int removedInitial) {
+      cdl = new CountDownLatch(nonChangeInitial);
+      cdls = new EnumMap<>(DocumentChange.Type.class);
+      cdls.put(DocumentChange.Type.ADDED, new CountDownLatch(addedInitial));
+      cdls.put(DocumentChange.Type.MODIFIED, new CountDownLatch(modifiedInitial));
+      cdls.put(DocumentChange.Type.REMOVED, new CountDownLatch(removedInitial));
```
schmidt-sebastian: The way the Watch is used in this test (adding or modifying a document at a time) is actually deterministic. By splitting these updates across types, you add some complexity to the tests that I don't think you need.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: Yeah, I'll clean those up.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: This type of scenario is specifically called out in the style guide: http://go/java-style#s5.2.3-method-names
> Underscores may appear in JUnit test method names to separate logical components of the name, with each component written in lowerCamelCase

Is your ask here for me to not use the upper enum names? i.e. `s/ADDED/added/`?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: What fix would you like me to apply here? Turn this into a sentence?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: Will do

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: I'll rename

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: I see what you're saying, but I'd prefer to keep this here for test reliability sake. There is a difference between "the listener expects an event within 5 seconds from the time the write is complete" to "the listener expects an event within 5 seconds from the time the write is sent". And while the time window there should be small, sometimes events are a bit slow being fired. I've seen a few runs while working on these tests fail in just this small time difference.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: Will update

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: I'll make sure all tests wait for the initial event

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: Will fix

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: I tried removing the individual CountDownLatches here and ran into several tests becoming more flaky, so I'd like to leave this for now and spend more time digging into it later.

BenWhitehead:
@schmidt-sebastian I've pushed several commits that have addressed most of your comments. (I added some more debugging information to the assertions in the case that they fail to hopefully aid in diagnostics when we see failures. I've seen a few failures working on things, seemingly related to events taking longer than we allow the `await` to block. Is there a defined SLO for events being delivered?)

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,550 @@
...
+  @AfterClass
+  public static void afterClass() throws Exception {
+    firestore.close();
+  }
+
+  /*
+  Attach a listener to a query with empty results.
+  Verify the listener receives an empty event.
+   */
+  @Test
```
schmidt-sebastian: Unless this is common in GCloud Java, may I recommend changing this to:


```
  /**
   * Attach a listener to a query with empty results.
   * Verify the listener receives an empty event.
   */
```

If you don't think is is JSDoc worthy then I would move it into the method.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,550 @@
...
+    QuerySnapshotEventListener listener = new QuerySnapshotEventListener(1);
+    ListenerRegistration registration = query.addSnapshotListener(listener);
+
+    try {
+      listener.eventsCountDownLatch.await();
+    } finally {
+      registration.remove();
+    }
+
+    ListenerAssertions la = listener.assertions();
```
schmidt-sebastian: Nit: No acronyms.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,550 @@
...
+  Attach a listener to a query with empty results.
+  Create a new document that matches the query.
+  Verify newly created document results in an ADDED event.
+  */
+  @Test
+  public void emptyResults_newDocument_ADDED()
+      throws InterruptedException, TimeoutException, ExecutionException {
+
+    final Query query = randomColl.whereEqualTo("foo", "bar");
+    QuerySnapshotEventListener listener = new QuerySnapshotEventListener(1, 1, 0, 0);
```
schmidt-sebastian: Can you add argument name comments to explain these constants?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
schmidt-sebastian: Sounds good.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,550 @@
...
+    }
+  }
+
+  private static final class EventsCountDownLatch {
+    private final CountDownLatch cdl;
+    private final EnumMap<DocumentChange.Type, CountDownLatch> cdls;
+
+    EventsCountDownLatch(
+        int nonChangeInitial, int addedInitial, int modifiedInitial, int removedInitial) {
+      cdl = new CountDownLatch(nonChangeInitial);
```
schmidt-sebastian: Nit: No acronyms.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,550 @@
...
+      this.eventsCountDownLatch =
+          new EventsCountDownLatch(nonChangeInitial, addedInitial, modifiedInitial, removedInitial);
+    }
+
+    @Override
+    public void onEvent(@Nullable QuerySnapshot value, @Nullable FirestoreException error) {
+      receivedEvents.add(new ListenerEvent(value, error));
+      if (value != null) {
+        List<DocumentChange> documentChanges = value.getDocumentChanges();
+        for (DocumentChange dc : documentChanges) {
```
schmidt-sebastian: Nit: No acronyms.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,550 @@
...
+    }
+
+    ListenerAssertions assertions() {
+      return new ListenerAssertions(receivedEvents);
+    }
+
+    static final class ListenerAssertions {
+      private static final MapJoiner MAP_JOINER = Joiner.on(",").withKeyValueSeparator("=");
+      final List<ListenerEvent> receivedEvents;
+      private final FluentIterable<ListenerEvent> events;
```
schmidt-sebastian: Do we need both receivedEvents and events?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,550 @@
...
+          builder.append(",");
+          builder.append("value=");
+          debugMessage(builder, receivedEvent.value);
+          builder.append("},\n");
+        }
+        builder.append("]");
+        return builder.toString();
+      }
+
+      private static void debugMessage(StringBuilder builder, QuerySnapshot qs) {
```
schmidt-sebastian: Nit: No acronyms.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: Nope, leftovers from refactorings. Cleaned up.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: Will do

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
BenWhitehead: I've refactored to a builder so we get the names with the arguments.

BenWhitehead:
@schmidt-sebastian I've taken care of all the nits, and I update the await conditions to be a multiple of the number of events rather than the firm 5 seconds. This will at least scale our tolerance in the test since.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryWatchTest.java
```diff
@@ -0,0 +1,641 @@
...
+
+  @AfterClass
+  public static void afterClass() throws Exception {
+    firestore.close();
+  }
+
+  /**
+   *
+   *
+   * <ol>
```
schmidt-sebastian: Nit: Are the two empty lines above from code formatting?

BenWhitehead: They are indeed. I wasn't able to figure out a way to remove them.

PR #108, opened 2020-02-14
fix: add cause to transaction errors on transaction commit

rockwotj:
Right now if a transaction fails due to a precondition failure (which are retried, which is another bug) for example trying to update a document that doesn't exist you get a bad error message of "too many retries". Ideally retry is smarter, but this is a easy change that propagates the underlying error up to the ApiFuture result.

/cc @schmidt-sebastian @BenWhitehead

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreException.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreException.java
@@ -28,7 +28,11 @@ public final class FirestoreException extends BaseGrpcServiceException {
   private Status status;
 
   private FirestoreException(String reason, Status status) {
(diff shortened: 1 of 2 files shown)


BenWhitehead:
@rockwotj Can you run `mvn com.coveo:fmt-maven-plugin:format` to get the code format pre-submit to pass?

schmidt-sebastian:
FYI: We will shortly be addressing the transaction retry behavior across all of our SDKs.

schmidt-sebastian:
@rockwotj Can you update the PR title to follow conventionalcommits.org?

It this case it would be: "fix: ddd cause to transaction errors on transaction commit"

BenWhitehead:
Thanks for the improvement @rockwotj!

PR #121, opened 2020-02-28
feat: new methods to get array and map fields without unchecked

### LINKED ISSUE #60: New methods to get Array and Map fields without unchecked casts
**Is your feature request related to a problem? Please describe.**

The _DocumentSnapshot_ class provides methods to retrieve a field by Firestore type, with the exception of 2 types: Array and Map. Retrieving the content of an Array or Map field and assigning it to a variable requires to call `get` and to do an unchecked cast. 

**Describe the solution you'd like**

For instance, adding 2 new methods such as `public <T> List<T> getArray(@Nonnull String field)` and `public <T> Map<String, T> getMap(@Nonnull String field)`.

**Describe alternatives you've considered**

Making an unchecked cast or using get with the second parameter that accepts a value type class but still requires an unchecked cast.

**Additional context**
I guess there is maybe a reason for not providing such methods (i.e. values for Array and Map are heterogeneous in Firestore and you do not want unchecked cast in your client library?). In that case, a note in the documentation suggesting the recommended approach would be helpful.

suraj-qlogic:
Fixes #60

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
@@ -469,6 +469,42 @@ public class DocumentSnapshot {
   }
 
   /**
+   * Returns the value of the field as an array.
+   *
+   * @param field The path to the field.
+   * @throws IllegalArgumentException if the value is not an array.
+   * @return The value of the field as an array.
+   */
+  @Nullable
+  public <T> List<T> getArray(@Nonnull String field) {
+    if (field == null) {
+      return null;
+    }
+    Object fieldValue = get(field);
+    Preconditions.checkArgument(
+        fieldValue instanceof List, "The field value at  " + field + " is not an array.");
+    return (List<T>) fieldValue;
+  }
+
+  /**
+   * Returns the value of the field as a Map.
+   *
+   * @param field The path to the field.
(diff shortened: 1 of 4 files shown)


BenWhitehead:
@suraj-qlogic Thanks for posting, I'm looking into some of our other libraries around how this type of scenario is handled. I'll report back when I've learned more.

As-is I'm a little wary to merge this as it's quite easy to cause a `ClassCastException`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
```diff
@@ -468,6 +468,42 @@ boolean isEmpty() {
...
+   * @throws IllegalArgumentException if the value is not an array.
+   * @return The value of the field as an array.
+   */
+  @Nullable
+  public <T> List<T> getArray(@Nonnull String field) {
+    if (field == null) {
+      return null;
+    }
+    Object fieldValue = get(field);
+    Preconditions.checkArgument(
```
elharo: I'm not sure this is the right exception here. The problem is perhaps not the argument.

BenWhitehead:
I spoke with several colleagues about this feature and we have decided that it it too risky to add to the library. As things stand right now it is far too easy for a ClassCastException to happen to a user, and we'd prefer to keep the compiler warning visible to users as they are the ones most familiar with their data. 

I will add a comment to the original issue as well outlining the decision.

PR #129, opened 2020-03-11
feat: base transaction retries on error codes     

schmidt-sebastian:
Port of https://github.com/googleapis/nodejs-firestore/pull/953

Implements go/transaction-retry-matrix for Java and cleans up some of the tests:
- Transaction IDs are passed as string to the helpers
- Request/response validation is centralized


schmidt-sebastian:
@BenWhitehead Can you ask someone to review or review this yourself if you have time? This one is a little trickier :/

BenWhitehead:
@schmidt-sebastian Sure, I'll dig into this tomorrow first thing.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
```diff
@@ -0,0 +1,279 @@
...
+          }
+        });
+    return callbackResult;
+  }
+
+  /** Determines whether the provided error is considered retryable. */
+  private boolean isRetryableTransactionError(ApiException exception) {
+    switch (exception.getStatusCode().getCode()) {
+        // This list is based on
+        // https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore/src/core/transaction_runner.ts#L112
```
BenWhitehead: Can you change this to a link with a SHA in it so that the line number is stable in the future?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
```diff
@@ -0,0 +1,279 @@
...
+  private Transaction transaction;
+  private int attemptsRemaining;
+
+  /**
+   * @param firestore The active Firestore instance
+   * @param userCallback The user provided transaction callback
+   * @param userCallbackExecutor The executor to run the user callback on
+   * @param numberOfAttempts The total number of attempts for this transaction
+   */
+  public TransactionRunner(
```
BenWhitehead: This can be package-private.

```suggestion
  TransactionRunner(
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
```diff
@@ -0,0 +1,279 @@
...
+import java.util.List;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Implements backoff and retry semantics for Firestore transactions.
+ *
+ * <p>Consumers should invoke `run()` to obtain an ApiFuture that resolves when all retries
+ * complete. Backoff settings can be customized via FirebaseOptions.
```
BenWhitehead: Looking at the constructor and use-site of this new class it appears `TransactionOptions` have some impact to settings here as well. If I'm reading this correctly can you also add it to the comment here.

Can you also add a note to this javadoc calling out which settings affect things here?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
```diff
@@ -0,0 +1,279 @@
...
+ */
+class TransactionRunner<T> {
+
+  private static final Tracer tracer = Tracing.getTracer();
+  private static final io.opencensus.trace.Status TOO_MANY_RETRIES_STATUS =
+      io.opencensus.trace.Status.ABORTED.withDescription("too many retries");
```
BenWhitehead: nit: The description here has spaces in it where as the failed one below has underscores

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
```diff
@@ -0,0 +1,279 @@
...
+                  }
+                },
+                MoreExecutors.directExecutor());
+          }
+        });
+    return callbackResult;
+  }
+
+  /** Determines whether the provided error is considered retryable. */
+  private boolean isRetryableTransactionError(ApiException exception) {
```
BenWhitehead: Can we move this method down into `RestartTransactionCallback` since it's the only thing calling this method?

(nit: This method can also be made static)

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -73,14 +73,23 @@
...
+          .setRetryDelayMultiplier(1)
+          .setJittered(false)
+          .build();
+
   public static final String DATABASE_NAME;
   public static final String COLLECTION_ID;
   public static final String DOCUMENT_PATH;
   public static final String DOCUMENT_NAME;
-  public static final ByteString TRANSACTION_ID;
+  public static final String TRANSACTION_ID;
```
BenWhitehead: What's the motivation for switching from `ByteString` -> `String` here and in the methods below?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
schmidt-sebastian: I expanded on the comment here and included both the constructor parameters as well as the backoff settings.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
schmidt-sebastian: "too many retries" existed before so I dropped the spaces from "user_callback_failed".

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
schmidt-sebastian: Fixed

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
schmidt-sebastian: Moved, but it cannot be static in the inner class anymore.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
schmidt-sebastian: Done

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
schmidt-sebastian: I moved the conversion to ByteString into the helper methods so that the API in this class accepts Strings.

PR #136, opened 2020-03-18
fix: add support for updating an individual field with pojo in all update method

### LINKED ISSUE #126: Firestore: support updating an individual field with pojo as a value from all update methods
The fix googleapis/google-cloud-java#6711 introduced an ability to update an individual field with pojo as a value. However, this does not work with all `update` method signatures.

`UpdateBuilder` has two private polymorphic `performUpdate` methods: the first one (1) that takes 5 arguments and internally calls `CustomClassMapper.convertToPlainJavaTypes ` followed by the call to the second `performUpdate` method (2) that takes three arguments:

```
private T performUpdate(
      @Nonnull DocumentReference documentReference,
      @Nonnull Precondition options,
      @Nonnull FieldPath fieldPath,
      @Nullable Object value,
      Object[] moreFieldsAndValues
) {
    Object data = CustomClassMapper.convertToPlainJavaTypes(value);
    // ...
    return performUpdate(documentReference, fields, options);
}
```

```
private T performUpdate(
      @Nonnull DocumentReference documentReference,
      @Nonnull final Map<FieldPath, Object> fields,
      @Nonnull Precondition precondition
) {
    ///...
}
```

`UpdateBuilder` has six public polymorphic `update` methods, but only four of them (the ones that take `value` argument of type `Object`) call (1), the other two are calling (2). The same is true for the six `update` methods in `DocumentReference`.

Not only this is a bit confusing, I also believe it is quite common to update multiple fields with a pojo as a value in a batch or transaction. Therefore it should be assumed that one or more values of the `Map<String, Object> fields` parameter can be pojos. This way the two versions of `performUpdate` could be merged into one or the only method that calls (2) could be (1).

To sum up, depending on which `update` signature is used, attempting to update a field with pojo as a value may still result in an exception (valid at least for google-cloud-firestore 1.32.4):

```
com.google.cloud.firestore.FirestoreException: Cannot convert DummyPojo to Firestore Value
    at com.google.cloud.firestore.FirestoreException.invalidState (FirestoreException.java:55)
    at com.google.cloud.firestore.UserDataConverter.encodeValue (UserDataConverter.java:180)
    at com.google.cloud.firestore.UserDataConverter.encodeValue (UserDataConverter.java:163)
    at com.google.cloud.firestore.DocumentSnapshot.fromObject (DocumentSnapshot.java:89)
    at com.google.cloud.firestore.UpdateBuilder.performUpdate (UpdateBuilder.java:514)
    at com.google.cloud.firestore.UpdateBuilder.update (UpdateBuilder.java:347)
```

suraj-qlogic:
Fixes #126

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
@@ -87,12 +87,14 @@ public final class LocalFirestoreHelper {
   public static final Map<String, Object> SINGLE_FIELD_MAP;
   public static final SingleField SINGLE_FIELD_OBJECT;
   public static final Map<String, Value> SINGLE_FIELD_PROTO;
+  public static final Map<String, Value> SINGLE_POJO_PROTO;
   public static final DocumentSnapshot SINGLE_FIELD_SNAPSHOT;
   public static final Value SINGLE_FIELD_VALUE;
   public static final SingleField UPDATE_SINGLE_FIELD_OBJECT;
   public static final Map<String, Object> UPDATED_FIELD_MAP;
   public static final Map<String, Value> UPDATED_FIELD_PROTO;
   public static final Map<String, Value> UPDATED_SINGLE_FIELD_PROTO;
+  public static final Map<String, Value> UPDATED_POJO_PROTO;
 
   public static final Map<String, Float> SINGLE_FLOAT_MAP;
   public static final Map<String, Value> SINGLE_FLOAT_PROTO;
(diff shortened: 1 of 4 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -191,6 +198,31 @@ public void setShortValue(@Nullable Short shortValue) {
     }
   }
 
+  public static class FooModel {
```
schmidt-sebastian: Do you mind re-using the SingleField example Pojo that exists in this class?

schmidt-sebastian:
@BenWhitehead This looks good to me, but we should try to replace the FooModel usage with SingleField before merging.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
suraj-qlogic: Done

PR #145, opened 2020-03-23
chore: Migrate firestore synth.py to bazel

vam-google:
The changes in grpc stubs are caused by the gRPC upgrade from 1.10 (more than a year old) to 1.27 (same version which is used as runtime dependency)


vam-google:
@chingor13 PTAL

vam-google:
@BenWhitehead PTAL

vam-google:
@BenWhitehead The binary compatibility works now, but java 7 and 11 fail for some reason. PTAL

PR #154, opened 2020-03-26
fix: mark v1beta1 as deprecated

schmidt-sebastian:
We would like to discourage use of the v1beta1 protocol.

Port of googleapis/nodejs-firestore#937


schmidt-sebastian:
cc @BenWhitehead

PR #156, opened 2020-03-27
feat: use SecureRandom instead of Random to reduce the chance of auto-id collisions

### LINKED ISSUE #155: Use SecureRandom instead of Random to reduce the chance of auto-id collisions
To reduce the likelihood of automatically-generated document ID collisions, the usage of `java.util.Random` should be upgraded to the more cryptographically-secure random number generator `java.security.SecureRandom`.

FYI, this change has already been made to the Android SDK: https://github.com/firebase/firebase-android-sdk/pull/1250

dconeybe:
- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-firestore/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #155 ☕️

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -41,6 +41,7 @@ import io.opencensus.trace.AttributeValue;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
 import io.opencensus.trace.Tracing;
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -57,7 +58,7 @@ import javax.annotation.Nullable;
  */
 class FirestoreImpl implements Firestore {
 
-  private static final Random RANDOM = new Random();
+  private static final Random RANDOM = new SecureRandom();
   private static final int AUTO_ID_LENGTH = 20;
   private static final String AUTO_ID_ALPHABET =
       "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";


BenWhitehead:
Do we know if users have ran into any issues with this sort of switch due to SecureRandom blocking to return values until there is enough entropy? 

> Note: Depending on the implementation, the generateSeed and nextBytes methods may block as entropy is being gathered, for example, if they need to read from /dev/random on various Unix-like operating systems. [1]

[1] https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html

dconeybe:
@BenWhitehead I do not know if this would be an issue.  I'll look into it though.  I could see it potentially being an issue with the Android SDK (where the change to SecureRandom has already occurred), where blocking the main thread is strongly discouraged; however, for a server-side SDK I wouldn't anticipate it being an issue.

BenWhitehead:
The main location I'd worry about it in the server side SDK would be in some of our "lean" environments, i.e. Cloud Functions, App Engine etc

BenWhitehead:
I've sent a question to several folks on my team to see if they've ran into anything around entropy in those environments and will report back if I hear anything.

dconeybe:
@BenWhitehead: @wilhuff pointed out that SecureRandom will not block in the way that it is used in this PR: https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/.  At least, it won't block on Mac or Linux.  In Windows, it's unclear if it will block or not.  But do those "lean environments" you referred to include Windows machines?

BenWhitehead:
I believe all the lean environments are linux/container based so `/dev/urandom` is probably a safe assumption, I'll go ahead an approve and merge this.

PR #211, opened 2020-05-09
fix: support array of references for IN queries

thebrianchen:
Porting from [node](https://github.com/googleapis/nodejs-firestore/pull/993).

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
@@ -717,7 +717,7 @@ public class Query {
                 "Invalid query. You cannot perform '%s' queries on FieldPath.documentId().",
                 operator.toString()));
       } else if (operator == IN) {
-        if (!(value instanceof List) || ((List<?>) value).size() == 0) {
+        if (!(value instanceof List) || ((List<?>) value).isEmpty()) {
           throw new IllegalArgumentException(
               String.format(
                   "Invalid Query. A non-empty array is required for '%s' filters.",
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
@@ -407,8 +407,10 @@ public class Query {
       reference = (DocumentReference) fieldValue;
     } else {
       throw new IllegalArgumentException(
-          "The corresponding value for FieldPath.documentId() must be a String or a "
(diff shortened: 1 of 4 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -709,7 +711,27 @@ public Query whereIn(@Nonnull FieldPath fieldPath, @Nonnull List<? extends Objec
...
       FieldPath fieldPath, StructuredQuery.FieldFilter.Operator operator, Object value) {
     if (fieldPath.equals(FieldPath.DOCUMENT_ID)) {
-      value = this.convertReference(value);
+      if (operator == ARRAY_CONTAINS || operator == ARRAY_CONTAINS_ANY) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Invalid query. You cannot perform '%s' queries on FieldPath.documentId().",
+                operator.toString()));
+      } else if (operator == IN) {
+        if (!(value instanceof List) || ((List<?>) value).size() == 0) {
```
schmidt-sebastian, line 720: Nit: Prefer `isEmpty()` as it is guaranteed to run in `O(1)`.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
```diff
@@ -273,6 +274,87 @@ public void withFieldPathFilter() throws Exception {
...
+    } catch (IllegalArgumentException e) {
+      assertEquals(
+          "The corresponding value for FieldPath.documentId() must be a String or a "
+              + "DocumentReference, but was: 42.",
+          e.getMessage());
+    }
+
+    try {
+      query.whereIn(FieldPath.documentId(), Arrays.<Object>asList(42)).get();
+      fail();
```
schmidt-sebastian, line 320: The Node tests uses "42", not "Arrays.<Object>asList(42)", which tests a different case. I think you are re-testing the case from right above here.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
thebrianchen: done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
thebrianchen: Removed test, since passing in "42" causes compiler error.

PR #216, opened 2020-05-13
fix: Fields used in whereIn should be equality filters

wu-hui:
Fields appear in `whereIn` queries are used as implicit orderBys. This is wrong because `In` should be an equality filter.

This change brings Java SDK in line with Node in terms of how equality/inequality filters are determined.

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
@@ -304,6 +304,39 @@ public class QueryTest {
   }
 
   @Test
+  public void inQueriesFieldsNotUsedInOrderBy() throws Exception {
+    doAnswer(queryResponse())
+        .when(firestoreMock)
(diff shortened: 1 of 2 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
```diff
@@ -303,6 +303,39 @@ public void inQueriesWithReferenceArray() throws Exception {
     assertEquals(expectedRequest, runQuery.getValue());
   }
 
+  @Test
+  public void inQueriesFieldsNotUsedInOrderBy() throws Exception {
```
wilhuff, line 307: You may also want to validate that unary filters are handled properly and that inequalities result in an implicit order by if that's not already present.

wu-hui: I think unary filters are basically isNull and isNan, which are equality checks by nature.

PR #230, opened 2020-05-27
fix: add RateLimiter

thebrianchen:
Porting from [node](https://github.com/googleapis/nodejs-firestore/pull/1021).

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
@@ -16,6 +16,7 @@
 
 package com.google.cloud.firestore;
 
+import com.google.common.base.Preconditions;
 import java.util.Date;
 
 /**
@@ -29,9 +30,10 @@ import java.util.Date;
  * <p>RateLimiter can also implement a gradually increasing rate limit. This is used to enforce the
  * 500/50/5 rule.
  *
- * @see "https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic"
+ * @see <a href=https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic>Ramping up
+ *     traffic</a>
  */
-public class RateLimiter {
+class RateLimiter {
   private final int initialCapacity;
   private final double multiplier;
   private final int multiplierMillis;
@@ -115,17 +117,15 @@ public class RateLimiter {
    *     testing the limiter.
    */
   private void refillTokens(long requestTimeMillis) {
-    if (requestTimeMillis >= lastRefillTimeMillis) {
-      long elapsedTime = requestTimeMillis - lastRefillTimeMillis;
-      int capacity = calculateCapacity(requestTimeMillis);
-      int tokensToAdd = (int) ((elapsedTime * capacity) / 1000);
-      if (tokensToAdd > 0) {
-        availableTokens = Math.min(capacity, availableTokens + tokensToAdd);
-        lastRefillTimeMillis = requestTimeMillis;
-      }
-    } else {
-      throw new IllegalArgumentException(
-          "Request time should not be before the last token refill time");
+    Preconditions.checkArgument(
+        requestTimeMillis >= lastRefillTimeMillis,
+        "Request time should not be before the last token refill time");
+    long elapsedTime = requestTimeMillis - lastRefillTimeMillis;
+    int capacity = calculateCapacity(requestTimeMillis);
+    int tokensToAdd = (int) ((elapsedTime * capacity) / 1000);
+    if (tokensToAdd > 0) {
+      availableTokens = Math.min(capacity, availableTokens + tokensToAdd);
+      lastRefillTimeMillis = requestTimeMillis;
     }
   }
 
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore;
+
+import java.util.Date;
+
+/**
+ * A helper that uses the Token Bucket algorithm to rate limit the number of operations that can be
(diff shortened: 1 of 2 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
```diff
@@ -0,0 +1,139 @@
...
+ * made in a second.
+ *
+ * <p>Before a given request containing a number of operations can proceed, RateLimiter determines
+ * doing so stays under the provided rate limits. It can also determine how much time is required
+ * before a request can be made.
+ *
+ * <p>RateLimiter can also implement a gradually increasing rate limit. This is used to enforce the
+ * 500/50/5 rule (https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic).
+ */
+public class RateLimiter {
```
schmidt-sebastian, line 32: There is already a RateLimiter in our classpath. Did you take a look if we can use it?

https://guava.dev/releases/19.0/api/docs/index.html?com/google/common/util/concurrent/RateLimiter.html

It supports a warm up period, but I am not sure if it matches our exact needs.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * A helper that uses the Token Bucket algorithm to rate limit the number of operations that can be
+ * made in a second.
+ *
+ * <p>Before a given request containing a number of operations can proceed, RateLimiter determines
+ * doing so stays under the provided rate limits. It can also determine how much time is required
+ * before a request can be made.
+ *
+ * <p>RateLimiter can also implement a gradually increasing rate limit. This is used to enforce the
+ * 500/50/5 rule (https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic).
```
schmidt-sebastian, line 30: Use `@link` JavaDoc here.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
```diff
@@ -0,0 +1,139 @@
...
+ */
+public class RateLimiter {
+  final int initialCapacity;
+  final double multiplier;
+  final int multiplierMillis;
+  final long startTimeMillis;
```
schmidt-sebastian, line 36: Can these be private?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
```diff
@@ -0,0 +1,139 @@
...
+   * the tokens were refilled.
+   *
+   * @param requestTimeMillis The time used to calculate the number of available tokens. Used for
+   *     testing the limiter.
+   */
+  private void refillTokens(long requestTimeMillis) {
+    if (requestTimeMillis >= lastRefillTimeMillis) {
+      long elapsedTime = requestTimeMillis - lastRefillTimeMillis;
+      int capacity = calculateCapacity(requestTimeMillis);
+      int tokensToAdd = (int) Math.floor((elapsedTime * capacity) / 1000);
```
schmidt-sebastian, line 119: You can just cast to "int". It does Math.floor() for you.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
```diff
@@ -0,0 +1,139 @@
...
+      if (tokensToAdd > 0) {
+        availableTokens = Math.min(capacity, availableTokens + tokensToAdd);
+        lastRefillTimeMillis = requestTimeMillis;
+      }
+    } else {
+      throw new IllegalArgumentException(
```
schmidt-sebastian, line 125: Optional: You Preconditions.checkArgument as first statement of this method.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
```diff
@@ -0,0 +1,139 @@
...
+          "Request time should not be before the last token refill time");
+    }
+  }
+
+  public int calculateCapacity(long requestTimeMillis) {
+    long millisElapsed = requestTimeMillis - startTimeMillis;
+    int operationsPerSecond =
+        (int)
```
schmidt-sebastian, line 133: Same comment regarding Math.floor + cast.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
thebrianchen: Is there any benefit to using `Preconditions.checkArgument` instead of throwing it here? Otherwise keeping it as is would mirror node.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
thebrianchen: done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
schmidt-sebastian: I was hoping the ramp up provided by this Google provided class would support our own network traffic guidance, but I also have my doubts. FWIW, I do think that at least one of the libraries we port to might have an adequate rate limiting algorithm already, and showing how to use one via an example would also have been helpful.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
schmidt-sebastian: Benefits:
- It is more Java idiomatic
- It removes the indent
- It will probably remove two lines from this method

With that being said, totally optional to change this.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
schmidt-sebastian: I would drop the quotes and maybe make this an `<a href=>` link like in those examples (it feels like overkill though).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
schmidt-sebastian: Unrelated: This class should not be public. I assume we can just make this package-private.

thebrianchen: changed to package-private. Thanks for the explanation -- I had not considered the value of an example that uses existing rate limiting algorithms.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
thebrianchen: added href wth the text "Ramping up traffic"

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RateLimiter.java
thebrianchen: changed. thanks for writing it out! I did not realize that checkArgument pattern was a Java idiom.

PR #233, opened 2020-06-03
fix: Add BulkWriter

thebrianchen:
Porting from node (multiple commits).

TODO:
- Add POJO support.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
@@ -19,6 +19,8 @@ package com.google.cloud.firestore;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
 import com.google.api.core.SettableApiFuture;
+import com.google.cloud.firestore.UpdateBuilder.BatchState;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.collect.FluentIterable;
@@ -113,7 +115,7 @@ public class BulkWriter {
    *
    * @param documentReference A reference to the document to be created.
    * @param fields A map of the fields and values for the document.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> create(
@@ -129,7 +131,8 @@ public class BulkWriter {
    * Delete a document from the database.
    *
    * @param documentReference The DocumentReference to delete.
-   * @return An ApiFuture containing the result of the delete. Throws an error if the delete fails.
+   * @return An ApiFuture containing the result of the delete. Contains an error if the delete
+   *     fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> delete(@Nonnull DocumentReference documentReference) {
@@ -144,15 +147,16 @@ public class BulkWriter {
    * Delete a document from the database.
    *
    * @param documentReference The DocumentReference to delete.
-   * @param options Preconditions to enforce for this delete.
-   * @return An ApiFuture containing the result of the delete. Throws an error if the delete fails.
+   * @param precondition Precondition to enforce for this delete.
+   * @return An ApiFuture containing the result of the delete. Contains an error if the delete
+   *     fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> delete(
-      @Nonnull DocumentReference documentReference, @Nonnull Precondition options) {
+      @Nonnull DocumentReference documentReference, @Nonnull Precondition precondition) {
     verifyNotClosed();
     BulkCommitBatch bulkCommitBatch = getEligibleBatch(documentReference);
-    ApiFuture<WriteResult> future = bulkCommitBatch.delete(documentReference, options);
+    ApiFuture<WriteResult> future = bulkCommitBatch.delete(documentReference, precondition);
     sendReadyBatches();
     return future;
   }
@@ -163,7 +167,7 @@ public class BulkWriter {
    *
    * @param documentReference A reference to the document to be set.
    * @param fields A map of the fields and values for the document.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> set(
@@ -183,7 +187,7 @@ public class BulkWriter {
    * @param documentReference A reference to the document to be set.
    * @param fields A map of the fields and values for the document.
    * @param options An object to configure the set behavior.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> set(
@@ -208,7 +212,7 @@ public class BulkWriter {
    *
    * @param documentReference A reference to the document to be updated.
    * @param fields A map of the fields and values for the document.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> update(
@@ -231,17 +235,17 @@ public class BulkWriter {
    *
    * @param documentReference A reference to the document to be updated.
    * @param fields A map of the fields and values for the document.
-   * @param options Preconditions to enforce on this update.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @param precondition Precondition to enforce on this update.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> update(
       @Nonnull DocumentReference documentReference,
       @Nonnull Map<String, Object> fields,
-      Precondition options) {
+      Precondition precondition) {
     verifyNotClosed();
     BulkCommitBatch bulkCommitBatch = getEligibleBatch(documentReference);
-    ApiFuture<WriteResult> future = bulkCommitBatch.update(documentReference, fields, options);
+    ApiFuture<WriteResult> future = bulkCommitBatch.update(documentReference, fields, precondition);
     sendReadyBatches();
     return future;
   }
@@ -259,7 +263,7 @@ public class BulkWriter {
    * @param field The first field to set.
    * @param value The first value to set.
    * @param moreFieldsAndValues String and Object pairs with more fields to be set.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> update(
@@ -288,7 +292,7 @@ public class BulkWriter {
    * @param fieldPath The first field to set.
    * @param value The first value to set.
    * @param moreFieldsAndValues String and Object pairs with more fields to be set.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> update(
@@ -317,19 +321,19 @@ public class BulkWriter {
    * @param field The first field to set.
    * @param value The first value to set.
    * @param moreFieldsAndValues String and Object pairs with more fields to be set.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> update(
       @Nonnull DocumentReference documentReference,
-      @Nonnull Precondition options,
+      @Nonnull Precondition precondition,
       @Nonnull String field,
       @Nullable Object value,
       Object... moreFieldsAndValues) {
     verifyNotClosed();
     BulkCommitBatch bulkCommitBatch = getEligibleBatch(documentReference);
     ApiFuture<WriteResult> future =
-        bulkCommitBatch.update(documentReference, options, field, value, moreFieldsAndValues);
+        bulkCommitBatch.update(documentReference, precondition, field, value, moreFieldsAndValues);
     sendReadyBatches();
     return future;
   }
@@ -344,23 +348,24 @@ public class BulkWriter {
    * by providing FieldPath objects.
    *
    * @param documentReference A reference to the document to be updated.
-   * @param options Preconditions to enforce on this update.
+   * @param precondition Precondition to enforce on this update.
    * @param fieldPath The first field to set.
    * @param value The first value to set.
    * @param moreFieldsAndValues String and Object pairs with more fields to be set.
-   * @return An ApiFuture containing the result of the write. Throws an error if the write fails.
+   * @return An ApiFuture containing the result of the write. Contains an error if the write fails.
    */
   @Nonnull
   public ApiFuture<WriteResult> update(
       @Nonnull DocumentReference documentReference,
-      @Nonnull Precondition options,
+      @Nonnull Precondition precondition,
       @Nonnull FieldPath fieldPath,
       @Nullable Object value,
       Object... moreFieldsAndValues) {
     verifyNotClosed();
     BulkCommitBatch bulkCommitBatch = getEligibleBatch(documentReference);
     ApiFuture<WriteResult> future =
-        bulkCommitBatch.update(documentReference, options, fieldPath, value, moreFieldsAndValues);
+        bulkCommitBatch.update(
+            documentReference, precondition, fieldPath, value, moreFieldsAndValues);
     sendReadyBatches();
     return future;
   }
@@ -369,7 +374,7 @@ public class BulkWriter {
    * Commits all writes that have been enqueued up to this point in parallel.
    *
    * <p>Returns an ApiFuture that completes when all currently queued operations have been
-   * committed. The ApiFuture will never throw an error since the results for each individual
+   * committed. The ApiFuture will never return an error since the results for each individual
    * operation are conveyed via their individual ApiFutures.
    *
    * <p>The ApiFuture completes immediately if there are no pending writes. Otherwise, the ApiFuture
@@ -402,7 +407,7 @@ public class BulkWriter {
   /**
    * Commits all enqueued writes and marks the BulkWriter instance as closed.
    *
-   * <p>After calling `close()`, calling any method wil throw an error.
+   * <p>After calling `close()`, calling any method wil return an error.
    *
    * <p>Returns an ApiFuture that completes when there are no more pending writes. The ApiFuture
    * will never error. Calling this method will send all requests. The ApiFuture completes
@@ -432,7 +437,7 @@ public class BulkWriter {
     if (batchQueue.size() > 0) {
       BulkCommitBatch lastBatch = batchQueue.get(batchQueue.size() - 1);
       if (lastBatch.getState() == UpdateBuilder.BatchState.OPEN
-          && !lastBatch.getDocPaths().contains(documentReference.getPath())) {
+          && !lastBatch.getDocuments().contains(documentReference)) {
         return lastBatch;
       }
     }
@@ -478,8 +483,8 @@ public class BulkWriter {
 
       // Send the batch if it is under the rate limit, or schedule another attempt after the
       // appropriate timeout.
-      long delayMs = rateLimiter.getNextRequestDelayMs(batch.getOpCount());
-      Preconditions.checkArgument(delayMs != -1, "Batch size should be under capacity");
+      long delayMs = rateLimiter.getNextRequestDelayMs(batch.getOperationCount());
+      Preconditions.checkState(delayMs != -1, "Batch size should be under capacity");
       if (delayMs == 0) {
         sendBatch(batch);
       } else {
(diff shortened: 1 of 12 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOptions.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.google.cloud.firestore;
+
+import javax.annotation.Nonnull;
+
+/** Options used to disable request throttling in BulkWriter. */
+public class BulkWriterOptions {
+  private final boolean disableThrottling;
+
+  BulkWriterOptions(boolean disableThrottling) {
```
thebrianchen, line 25: I noticed that all the other options objects have private constructors and expose `create()` instead. Should I do this here as well?

schmidt-sebastian: The idea behind these classes is to offer "named arguments". In this case, we would want to have a static factory method that is called "disableThrottling" or "withThrottlingDisabled". I would suggest to look at SetOptions rather than at TransactionOptions for inspiration :)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -94,6 +94,18 @@ public WriteBatch batch() {
     return new WriteBatch(this);
   }
 
+  @Nonnull
+  @Override
+  public BulkWriter bulkWriter() {
+    return new BulkWriter(this, true);
```
schmidt-sebastian, line 100: Add inline comment that describes `true`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -47,17 +49,47 @@
  * Abstract class that collects and bundles all write operations for {@link Transaction} and {@link
  * WriteBatch}.
  */
-public abstract class UpdateBuilder<T extends UpdateBuilder> {
+public abstract class UpdateBuilder<T> {
```
schmidt-sebastian, line 52: This implementation is now very different from Node. Do you think we can easily get Node to match? I am a little hesitant to provide two drastically different reference implementations.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -47,17 +49,47 @@
...
+    this(firestore, BulkWriter.MAX_BATCH_SIZE);
+  }
+
+  UpdateBuilder(FirestoreImpl firestore, int maxBatchSize) {
     this.firestore = firestore;
+    this.maxBatchSize = maxBatchSize;
     this.writes = new ArrayList<>();
   }
 
+  abstract T wrapResult(ApiFuture<WriteResult> result);
```
schmidt-sebastian, line 86: Please add a method comment, especially since this is an abstract class.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -675,4 +712,67 @@ boolean isEmpty() {
   public int getMutationsSize() {
     return writes.size();
   }
+
+  BatchState getState() {
+    return state;
+  }
+
+  HashSet<String> getDocPaths() {
```
schmidt-sebastian, line 720: Please return `Set` as returning a HashSet ties us to this implementation.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -675,4 +712,67 @@ boolean isEmpty() {
...
+    return docPaths;
+  }
+
+  int getOpCount() {
```
schmidt-sebastian, line 724: Java doesn't use abbreviations.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -675,4 +712,67 @@ boolean isEmpty() {
...
+    if (getOpCount() == maxBatchSize) {
+      state = BatchState.READY_TO_SEND;
+    }
+
+    return result;
+  }
+  /** Resolves the individual operations in the batch with the results. */
+  void processResults(List<BatchWriteResult> results, @Nullable Exception error) {
+    if (error == null) {
+      for (int i = 0; i < getOpCount(); ++i) {
```
schmidt-sebastian, line 746: You are looping over the resultsMap, but this is not apparent. This would be easier to read if you inlined `getOpCount` here and below (and maybe we can remove it altogether?).

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -0,0 +1,568 @@
...
+    BatchWriteResponse.Builder response = BatchWriteResponse.newBuilder();
+    response.addWriteResultsBuilder().getUpdateTimeBuilder().setSeconds(seconds).build();
+    response.addStatusBuilder().build();
+    return ApiFutures.immediateFuture(response.build());
+  }
+
+  private ApiFuture<BatchWriteResponse> failResponse() {
+    BatchWriteResponse.Builder response = BatchWriteResponse.newBuilder();
+    response.addWriteResultsBuilder().build();
+    response.addStatusBuilder().setCode(14).build();
```
schmidt-sebastian, line 89: Use the Code enum here?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -0,0 +1,568 @@
...
+  }
+
+  @Test
+  public void surfacesErrors() throws Exception {
+    RequestResponseMap requestResponseMap =
+        new RequestResponseMap() {
+          {
+            put(
+                batchWrite(set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc1")),
+                failResponse());
```
schmidt-sebastian, line 212: Should we pass the error code to `failResponse`? 

Nit: I think the name should be `failedResponse`, since it returns a failed response.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -0,0 +1,568 @@
...
+  }
+
+  @Test
+  public void addsWritesToNewBatchAfterFlush() throws Exception {
+    RequestResponseMap requestResponseMap =
+        new RequestResponseMap() {
+          {
+            put(
+                batchWrite(create(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc1")),
+                successResponse(2));
```
schmidt-sebastian, line 245: Use `1` here?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -0,0 +1,568 @@
...
+  }
+
+  @Test
+  public void sendsWritesToSameDocInSeparateBatches() throws Exception {
+    RequestResponseMap requestResponseMap =
+        new RequestResponseMap() {
+          {
+            put(
+                batchWrite(set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc1")),
+                successResponse(0));
```
schmidt-sebastian, line 320: IMHO 1 is a better starting point since it is different from the default timestamp.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -0,0 +1,568 @@
...
+    requestResponseMap.initializeStub(batchWriteCapture, firestoreMock);
+
+    bulkWriter.setMaxBatchSize(3);
+    ApiFuture<WriteResult> result1 = bulkWriter.set(doc1, LocalFirestoreHelper.SINGLE_FIELD_MAP);
+    ApiFuture<WriteResult> result2 = bulkWriter.update(doc2, LocalFirestoreHelper.SINGLE_FIELD_MAP);
+    ApiFuture<WriteResult> result3 =
+        bulkWriter.create(
+            firestoreMock.document("coll/doc3"), LocalFirestoreHelper.SINGLE_FIELD_MAP);
+
+    // The 4th write should not be sent because it should be in a new batch. However, this write
```
schmidt-sebastian, line 391: If I read the code correctly, the 3rd write sends the batch.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -0,0 +1,568 @@
...
+                  public ApiFuture<Void> call() {
+                    return bulkWriter.flush();
+                  }
+                },
+                0,
+                TimeUnit.MILLISECONDS);
+
+    // Wait for flush() to perform logic and reach the stubbed response. This simulates a first
+    // batch that has been sent with its response still pending.
+    requestResponseMap.awaitLatch();
```
schmidt-sebastian, line 438: Please use a more descriptive name for this API that makes it obvious what this latch does. Would `awaitRequest`/`awaitNextRequest` work?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -0,0 +1,568 @@
...
+                batchWrite(
+                    set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc1"),
+                    set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc2")),
+                mergeResponses(successResponse(1), successResponse(2)));
+            put(
+                batchWrite(set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc1")),
+                successResponse(3));
+          }
+        };
+    requestResponseMap.initializeStub(batchWriteCapture, firestoreMock, true);
```
schmidt-sebastian, line 419: What is `true`?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -0,0 +1,568 @@
...
+    verifyRequests(requests, requestResponseMap);
+  }
+
+  @Test
+  public void doesNotSendBatchesIfDoingSoExceedsRateLimit() throws Exception {
+    // The test is considered a success if BulkWriter tries to send the second batch again after a
+    // timeout.
+
+    List<Integer> arrayRange = new ArrayList<>();
+    for (int i = 0; i < 500; ++i) {
```
schmidt-sebastian, line 461: Can we "compute" this value inline?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -889,4 +928,72 @@ public static String autoId() {
   public static Map<String, Object> fromSingleQuotedString(String json) {
     return fromJsonString(json.replace("'", "\""));
   }
+
+  static class RequestResponseMap
```
schmidt-sebastian, line 932: I feel this should no longer be called `RequestResponseMap`, this does much more now. Can you come up with a more specific name?

I also think we could split this into two classes - the base RequestResponseMap and a subclass that enforces the one active request setting. With good naming (tm) it would now be super obvious what is going on.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -889,4 +928,72 @@ public static String autoId() {
...
+    SettableApiFuture<Void> activeRequestComplete = SettableApiFuture.create();
+    CountDownLatch latch = new CountDownLatch(1);
+
+    void markComplete() {
+      activeRequestComplete.set(null);
+    }
+
+    void awaitLatch() {
+      try {
+        latch.await();
```
schmidt-sebastian, line 944: You should try to generalized this code if you keep it in this generalized helper. A good way to do this would be a BlockingQueue of requests. This could be `awaitRequest` and use `BlockingQueue.poll()` to wait for the next request. You would supply the requests in the stub handler (and call `BlockingQueue.offer()`).

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -889,4 +928,72 @@ public static String autoId() {
...
     return fromJsonString(json.replace("'", "\""));
   }
+
+  static class RequestResponseMap
+      extends LinkedHashMap<GeneratedMessageV3, ApiFuture<? extends GeneratedMessageV3>> {
+    int activeRequestCounter = 0;
+    SettableApiFuture<Void> activeRequestComplete = SettableApiFuture.create();
+    CountDownLatch latch = new CountDownLatch(1);
+
+    void markComplete() {
```
schmidt-sebastian, line 938: Please use more descriptive name.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
thebrianchen: Does this modified implementation work? I'm still using a single latch and future, but it can process multiple serial requests.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOptions.java
schmidt-sebastian: Not yet.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
schmidt-sebastian: Let's live as is. Java remains outlier. We need to point that out in our docs for the contractors.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -929,22 +929,45 @@ public static String autoId() {
     return fromJsonString(json.replace("'", "\""));
   }
 
-  static class RequestResponseMap
+  /** Map of request/response pairs to be returned when `sendRequest()` is called. */
+  static class RequestResponseStubber
```
schmidt-sebastian, line 933: Nit: This only stubs responses. Maybe ResponseStubber?

It also seems like the comment is no longer applicable.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -929,22 +929,45 @@ public static String autoId() {
...
+      }
+    }
+  }
+
+  /**
+   * Map of request/response pairs to be returned when `sendRequest()` is called.
+   *
+   * <p>Enforces that only one active request can be pending at a time.
+   */
+  static class SerialRequestStubber extends RequestResponseStubber {
```
schmidt-sebastian, line 954: Similarly, this could be SerialResponseStubber.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -929,22 +929,45 @@ public static String autoId() {
...
     int activeRequestCounter = 0;
     SettableApiFuture<Void> activeRequestComplete = SettableApiFuture.create();
     CountDownLatch latch = new CountDownLatch(1);
 
-    void markComplete() {
+    void markRequestComplete() {
```
schmidt-sebastian, line 959: Maybe `markAllRequestsComplete()`? Not sure.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -929,22 +929,45 @@ public static String autoId() {
...
       activeRequestComplete.set(null);
     }
 
-    void awaitLatch() {
+    void awaitRequest() {
```
schmidt-sebastian, line 963: awaitNextRequest?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -929,22 +929,45 @@ public static String autoId() {
...
+      activeRequestComplete = SettableApiFuture.create();
       try {
         latch.await();
       } catch (Exception e) {
         fail("latch.await() should not fail");
       }
+      latch = new CountDownLatch(1);
```
schmidt-sebastian, line 970: I think you have to use semaphores here to allow for multiple requests to build up and be consumed. The state needs to span the lifetime of the Stubber.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOptions.java
```diff
@@ -0,0 +1,43 @@
...
+/** Options used to disable request throttling in BulkWriter. */
+public class BulkWriterOptions {
+
+  private final boolean disableThrottling;
+
+  private BulkWriterOptions(boolean disableThrottling) {
+    this.disableThrottling = disableThrottling;
+  }
+
+  boolean getDisableThrottling() {
```
schmidt-sebastian, line 30: This should be `isThrottlingEnabled`

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
schmidt-sebastian: I don't see why we need it. There is only one place where we are looping over `arrayRange` and we could just use `for (int i = 0; i < 500; ++i)` instead.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
schmidt-sebastian: If I follow this code correctly, you cannot wait for `n` requests or buffer `n` requests. You have to send, wait, send, wait, send and so on. If I send 2 requests, I should be able to invoke `awaitRequest()` twice without blocking.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -889,4 +928,84 @@ public static String autoId() {
...
+    void initializeStub(
+        ArgumentCaptor<? extends Message> argumentCaptor, FirestoreImpl firestoreMock) {
+      Stubber stubber = null;
+      for (final ApiFuture<? extends GeneratedMessageV3> response : values()) {
+        Answer<ApiFuture<? extends GeneratedMessageV3>> answer =
+            new Answer<ApiFuture<? extends GeneratedMessageV3>>() {
+              @Override
+              public ApiFuture<? extends GeneratedMessageV3> answer(
+                  InvocationOnMock invocationOnMock) throws Throwable {
+                return checkOneActiveRequestAndReturnResponse(response);
```
schmidt-sebastian, line 999: If you make checkOneActiveRequestAndReturnResponse() a validateRequest() method that in the base class doesn't do anything and in this class is `checkOneActiveRequestAndReturnResponse()`, then you can have a single shared implementation of `initializeStub()`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOptions.java
```diff
@@ -20,24 +20,24 @@
 
 /** Options used to disable request throttling in BulkWriter. */
 public class BulkWriterOptions {
+
   private final boolean disableThrottling;
 
-  BulkWriterOptions(boolean disableThrottling) {
+  private BulkWriterOptions(boolean disableThrottling) {
```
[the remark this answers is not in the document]
thebrianchen: done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
thebrianchen: I changed it from CountDownLatch to Semaphore, and I think you can buffer multiple requests now. Buffering `n` requests releases `n` permits that `n` `awaitRequest()` calls can instantly consume without blocking.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
thebrianchen: Great suggestion, thanks!

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
thebrianchen: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -0,0 +1,572 @@
...
+    }
+    sendReadyBatches();
+    ApiFutures.allAsList(writeFutures)
+        .addListener(
+            new Runnable() {
+              public void run() {
+                flushComplete.set(null);
+              }
+            },
+            MoreExecutors.directExecutor());
```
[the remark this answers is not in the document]
thebrianchen: Thanks for checking :)

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -929,14 +928,30 @@ public static String autoId() {
...
 
-  /** Map of request/response pairs to be returned when `sendRequest()` is called. */
-  static class RequestResponseStubber
+  /**
+   * Contains a map of request/response pairs that are used to create stub responses when
+   * `sendRequest()` is called.
+   */
+  static class ResponseStubber
       extends LinkedHashMap<GeneratedMessageV3, ApiFuture<? extends GeneratedMessageV3>> {
+    ApiFuture<? extends GeneratedMessageV3> checkOneActiveRequestAndReturnResponse(
```
schmidt-sebastian, line 937: This method should:
- have a generic name
- be document
- be overridden in the SerialResponseStubber and perform a "checkOneActiveRequestAndReturnResponse".

schmidt-sebastian: (ignore the previous version of this comment)

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -889,4 +927,79 @@ public static String autoId() {
...
+            new Answer<ApiFuture<? extends GeneratedMessageV3>>() {
+              @Override
+              public ApiFuture<? extends GeneratedMessageV3> answer(
+                  InvocationOnMock invocationOnMock) throws Throwable {
+                return checkOneActiveRequestAndReturnResponse(response);
+              }
+            };
+        stubber = (stubber != null) ? stubber.doAnswer(answer) : doAnswer(answer);
+      }
+      if (stubber != null) {
```
schmidt-sebastian, line 956: If stubber is null here, how would the tests work? Seems like we should assert that `stubber` is not null.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
thebrianchen: Renamed to `verifyResponse` and added comments.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
thebrianchen: Good point, the if statement is a lazy solution. Added a `checkNotNull`

PR #241, opened 2020-06-05
feat: add API to serialize Queries to Proto

schmidt-sebastian:
This PR adds an API to serialize and deserialize Queries to and from RunQueryRequests. The new API endpoints are `Query.toProto` and `Query.fromProto`.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
@@ -19,8 +19,11 @@ package com.google.cloud.firestore;
 import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
+import com.google.api.core.InternalExtensionOnly;
 import com.google.api.gax.rpc.ApiException;
 import com.google.api.gax.rpc.ApiExceptions;
+import com.google.api.gax.rpc.UnaryCallable;
+import com.google.cloud.firestore.spi.v1.FirestoreRpc;
 import com.google.cloud.firestore.v1.FirestoreClient.ListDocumentsPagedResponse;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.MoreExecutors;
@@ -40,16 +43,17 @@ import javax.annotation.Nullable;
  * test mocks. Subclassing is not supported in production code and new SDK releases may break code
  * that does so.
  */
+@InternalExtensionOnly
 public class CollectionReference extends Query {
 
   /**
    * Creates a CollectionReference from a complete collection path.
    *
-   * @param firestore The Firestore client.
+   * @param rpcContext The Firestore client.
    * @param collectionPath The Path of this collection.
    */
-  protected CollectionReference(FirestoreImpl firestore, ResourcePath collectionPath) {
-    super(firestore, collectionPath);
+  CollectionReference(FirestoreRpcContext<?> rpcContext, ResourcePath collectionPath) {
+    super(rpcContext, collectionPath);
   }
 
   /**
@@ -71,7 +75,7 @@ public class CollectionReference extends Query {
   @Nullable
   public DocumentReference getParent() {
     ResourcePath parent = options.getParentPath();
-    return parent.isDocument() ? new DocumentReference(firestore, parent) : null;
+    return parent.isDocument() ? new DocumentReference(rpcContext, parent) : null;
   }
 
   /**
@@ -109,7 +113,7 @@ public class CollectionReference extends Query {
     Preconditions.checkArgument(
         documentPath.isDocument(),
         String.format("Path should point to a Document Reference: %s", getPath()));
-    return new DocumentReference(firestore, documentPath);
+    return new DocumentReference(rpcContext, documentPath);
   }
 
   /**
@@ -133,10 +137,12 @@ public class CollectionReference extends Query {
     final ListDocumentsPagedResponse response;
 
     try {
-      response =
-          ApiExceptions.callAndTranslateApiException(
-              firestore.sendRequest(
-                  request.build(), firestore.getClient().listDocumentsPagedCallable()));
+      FirestoreRpc client = rpcContext.getClient();
+      UnaryCallable<ListDocumentsRequest, ListDocumentsPagedResponse> callable =
+          client.listDocumentsPagedCallable();
+      ListDocumentsRequest build = request.build();
+      ApiFuture<ListDocumentsPagedResponse> future = rpcContext.sendRequest(build, callable);
+      response = ApiExceptions.callAndTranslateApiException(future);
     } catch (ApiException exception) {
       throw FirestoreException.apiException(exception);
     }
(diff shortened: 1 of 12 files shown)


schmidt-sebastian:
@BenWhitehead This is now ready for review.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1162,6 +1186,113 @@ public void onCompleted() {
...
+  /**
+   * Returns the {@link RunQueryRequest} that this Query instance represents. The request contains
+   * the serialized form of all Query constraints.
+   *
+   * <p>Runtime metadata (as required for `limitToLast()` queries) is not serialized and as such,
+   * the serialized request will return the results in the original backend order.
+   *
+   * @return the serialized RunQueryRequest
+   */
+  public RunQueryRequest serializeToProto() {
```
wilhuff, line 1198: The internal convention seems to be `toProto` and `fromProto`, which seems perfectly fine. Why the extra long name here?

schmidt-sebastian: No strong opinion here. Renamed.

BenWhitehead:
I've just pushed a new commit that refactors things so we don't have to depend on FirestoreImpl as the input and allows for `Proxy`'d instances to satisfy our constraints.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
```diff
@@ -48,7 +52,7 @@
    * @param firestore The Firestore client.
    * @param collectionPath The Path of this collection.
    */
-  protected CollectionReference(FirestoreImpl firestore, ResourcePath collectionPath) {
+  CollectionReference(FirestoreRpcContext<?> firestore, ResourcePath collectionPath) {
```
schmidt-sebastian, line 55: Should the constructor be marked `@InternalExtensionOnly` rather than the whole class?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
```diff
@@ -42,13 +43,14 @@
...
+@InternalExtensionOnly
 public class DocumentReference {
 
   private final ResourcePath path;
-  private final FirestoreImpl firestore;
+  private final FirestoreRpcContext<?> firestore;
 
-  protected DocumentReference(
-      FirestoreImpl firestore, ResourcePath path) { // Elevated access level for mocking.
+  DocumentReference(
```
schmidt-sebastian, line 52: Same comment here (and likely everywhere).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -353,13 +359,19 @@ FirestoreRpc getClient() {
...
+  @Override
+  public <RequestT, ResponseT> ApiStreamObserver<RequestT> streamRequest(
       ApiStreamObserver<ResponseT> responseObserverT,
       BidiStreamingCallable<RequestT, ResponseT> callable) {
     Preconditions.checkState(!closed, "Firestore client has already been closed");
     return callable.bidiStreamingCall(responseObserverT);
   }
 
+  @Override
+  public FirestoreImpl getFirestore() {
```
schmidt-sebastian, line 371: I wish this was called differently, but I cannot think of a better name than the already taken `getClient()`. The frequent calls to `firestore.getFirestore()` seems slightly strange to me.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreRpcContext.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.api.core.InternalExtensionOnly;
+import com.google.api.gax.rpc.ApiStreamObserver;
+import com.google.api.gax.rpc.BidiStreamingCallable;
+import com.google.api.gax.rpc.ServerStreamingCallable;
+import com.google.api.gax.rpc.UnaryCallable;
+import com.google.cloud.firestore.spi.v1.FirestoreRpc;
+
+@InternalApi
+@InternalExtensionOnly
+interface FirestoreRpcContext<FS extends Firestore> {
```
schmidt-sebastian, line 30: My takeaway from our discussion last week that that this was intended to be public. Let me know if I misunderstood you :)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
BenWhitehead: Each of the model classes has the following note on it:

> <b>Subclassing Note</b>: Firestore classes are not meant to be subclassed except for use in
> test mocks. Subclassing is not supported in production code and new SDK releases may break code
> that does so.

Given that fact I've gone ahead and added the annotation to each of the classes touched here.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
BenWhitehead: See above for reasoning.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
schmidt-sebastian: Sounds good. I actually did my research this time:

```
Interfaces marked with @InternalExtensionOnly are public, but should only be implemented by internal classes. For the purposes of semver, we reserve the right to add to these interfaces without default implementations (for Java 7).
```

So yes, this should be on the class and not the constructor.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreRpcContext.java
BenWhitehead: I think that was my original intent, but after poking at things and seeing the amount of interconnectedness between the data and the operations I think we need to scope this to the level `FirestoreImpl` currently is. If this interface were as simple as `getDatabseName()` and `getResourcePath()` I think it'd be easier for us to have it public, but there is a lot of coupling against these methods that we don't want on the public api surface.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
BenWhitehead: Yeah, I'm not crazy about it either.

This method is serving two purposes I can see:
1. Direct access to the Firestore instance for return from a models `getFirestore()` method
2. Access to allow calling `Firestore#batch()` (14 times) or `Firestore#getAll(DocumentReference...)` (1 time)

Regarding 1 above, I don't think there is much we can do there since those `getFirestore()` are part of the public API. For 2, given the necessity in 1 I don't know that it makes sense for us too refactor things to remove the direct dependence on 2 at this time (especially with the upcoming work on BulkWriter).

would it help to rename the usage locations of the `FirestoreRpcContext` from `firestore` to something like `context` or `firestoreRpcContext`? (I didn't do this in my first push, to try and help reduce diff noise.)

schmidt-sebastian: I think `rpcContext` would be great and is as descriptive as `firestoreRpcContext` in the context of this repo.

BenWhitehead: I'll push a new commit with this change in just a few minutes.

BenWhitehead: New commit pushed

PR #286, opened 2020-07-07
fix: add retry on ABORTED errors

thebrianchen:
Porting over from [node](https://github.com/googleapis/nodejs-firestore/pull/1243).

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
@@ -29,7 +29,9 @@ public final class BatchWriteResult {
   @Nullable private final Exception exception;
 
   BatchWriteResult(
-      DocumentReference documentReference, @Nullable Timestamp timestamp, Exception exception) {
+      DocumentReference documentReference,
+      @Nullable Timestamp timestamp,
+      @Nullable Exception exception) {
     this.documentReference = documentReference;
     this.writeTime = timestamp;
     this.exception = exception;
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
@@ -17,7 +17,6 @@
 package com.google.cloud.firestore;
 
 import com.google.cloud.Timestamp;
-import io.grpc.Status;
 import javax.annotation.Nullable;
 
 /**
@@ -25,49 +24,24 @@ import javax.annotation.Nullable;
  * BatchWriteRequests.
  */
 public final class BatchWriteResult {
-  private final String key;
+  private final DocumentReference documentReference;
   @Nullable private final Timestamp writeTime;
-  private final Status status;
-  private final String message;
   @Nullable private final Exception exception;
 
-  BatchWriteResult(String key, @Nullable Timestamp timestamp, Status status, String message) {
-    this.key = key;
-    this.writeTime = timestamp;
-    this.status = status;
-    this.message = message;
-    this.exception = null;
-  }
-
   BatchWriteResult(
-      String key,
-      @Nullable Timestamp timestamp,
-      Status status,
-      String message,
-      Exception exception) {
-    this.key = key;
+      DocumentReference documentReference, @Nullable Timestamp timestamp, Exception exception) {
+    this.documentReference = documentReference;
     this.writeTime = timestamp;
-    this.status = status;
-    this.message = message;
     this.exception = exception;
   }
 
-  public String getKey() {
-    return key;
+  public DocumentReference getDocumentReference() {
+    return documentReference;
   }
 
   @Nullable
   public Timestamp getWriteTime() {
     return writeTime;
-  }
-
-  @Nullable
-  public Status getStatus() {
-    return status;
-  }
-
-  public String getMessage() {
-    return message;
   }
 
   @Nullable
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
@@ -25,14 +25,35 @@ import javax.annotation.Nullable;
  * BatchWriteRequests.
  */
 public final class BatchWriteResult {
+  private final String key;
   @Nullable private final Timestamp writeTime;
   private final Status status;
   private final String message;
+  @Nullable private final Exception exception;
 
-  BatchWriteResult(@Nullable Timestamp timestamp, Status status, String message) {
+  BatchWriteResult(String key, @Nullable Timestamp timestamp, Status status, String message) {
+    this.key = key;
     this.writeTime = timestamp;
     this.status = status;
     this.message = message;
+    this.exception = null;
+  }
+
+  BatchWriteResult(
+      String key,
+      @Nullable Timestamp timestamp,
+      Status status,
+      String message,
+      Exception exception) {
+    this.key = key;
+    this.writeTime = timestamp;
+    this.status = status;
+    this.message = message;
+    this.exception = exception;
+  }
+
+  public String getKey() {
+    return key;
   }
 
   @Nullable
@@ -40,11 +61,17 @@ public final class BatchWriteResult {
     return writeTime;
   }
 
+  @Nullable
   public Status getStatus() {
     return status;
   }
 
   public String getMessage() {
     return message;
+  }
+
+  @Nullable
+  public Exception getException() {
+    return exception;
   }
 }
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
@@ -32,6 +32,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -45,25 +46,24 @@ class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
     super(firestore, maxBatchSize);
   }
 
-  //  BulkCommitBatch(
-  //      FirestoreImpl firestore,
-  //      BulkCommitBatch retryBatch,
-  //      final Set<DocumentReference> docsToRetry) {
-  //    super(firestore);
-  //    this.writes.addAll(
-  //        FluentIterable.from(retryBatch.writes)
-  //            .filter(
-  //                new Predicate<WriteOperation>() {
-  //                  @Override
-  //                  public boolean apply(WriteOperation writeOperation) {
-  //                    return docsToRetry.contains(writeOperation.documentReference);
-  //                  }
-  //                })
-  //            .toList());
-  //
-  //    this.state = BatchState.READY_TO_SEND;
-  //    // TODO: figure out how to pipe completeFuture through into the new batch
-  //  }
+  BulkCommitBatch(
+      FirestoreImpl firestore,
+      BulkCommitBatch retryBatch,
+      final Set<DocumentReference> docsToRetry) {
+    super(firestore);
+    this.writes.addAll(
+        FluentIterable.from(retryBatch.writes)
+            .filter(
+                new Predicate<WriteOperation>() {
+                  @Override
+                  public boolean apply(WriteOperation writeOperation) {
+                    return docsToRetry.contains(writeOperation.documentReference);
+                  }
+                })
+            .toList());
+
+    this.state = retryBatch.state;
+  }
 
   ApiFuture<WriteResult> wrapResult(ApiFuture<WriteResult> result) {
     return result;
@@ -481,9 +481,10 @@ public class BulkWriter {
   public ApiFuture<Void> flush() {
     verifyNotClosed();
     final SettableApiFuture<Void> flushComplete = SettableApiFuture.create();
-    List<ApiFuture<Void>> writeFutures = new ArrayList<>();
+    List<SettableApiFuture<WriteResult>> writeFutures = new ArrayList<>();
     for (BulkCommitBatch batch : batchQueue) {
-      writeFutures.add(batch.awaitBulkCommit());
+      batch.markReadyToSend();
+      writeFutures.addAll(batch.getPendingFutures());
     }
     sendReadyBatches();
     ApiFutures.allAsList(writeFutures)
@@ -605,12 +606,13 @@ public class BulkWriter {
    * next group of ready batches.
    */
   private void sendBatch(final BulkCommitBatch batch) {
-    System.out.println("sending batch" + batch.getPendingDocuments().size());
+    Preconditions.checkState(
+        batch.state == BatchState.READY_TO_SEND,
+        "The batch should be marked as READY_TO_SEND before committing");
+    batch.state = BatchState.SENT;
     boolean success = rateLimiter.tryMakeRequest(batch.getPendingOperationCount());
     Preconditions.checkState(success, "Batch should be under rate limit to be sent.");
 
-    // Schedule the actual RPC call on Firestore's executor so that it does not block the main
-    // thread.
     ApiFuture<Void> commitFuture = bulkCommit(batch);
     commitFuture.addListener(
         new Runnable() {
@@ -622,9 +624,6 @@ public class BulkWriter {
             sendReadyBatches();
           }
         },
-        // TODO: Replacing this with the firestoreExecutor will result in tests passing, but batches
-        // being removed multiple times. Using the directExecutor results in the precondition check
-        // above failing when all tests are run together (run sendBatchesWhenSizeLimitIsReached)
         MoreExecutors.directExecutor());
   }
 
@@ -649,18 +648,15 @@ public class BulkWriter {
               String.format(
                   "Current batch failed at retry #%d. Num failures: %d",
                   attempt, batch.getPendingOperationCount()));
-          batch.sliceBatchForRetry(batch.getPendingDocuments());
-          batch.markReadyToSend();
 
           if (attempt < MAX_RETRY_ATTEMPTS) {
             nextAttempt = backoff.createNextAttempt(nextAttempt);
-            return bulkCommit(batch, attempt + 1);
+            BulkCommitBatch newBatch =
+                new BulkCommitBatch(firestore, batch, batch.getPendingDocuments());
+            return bulkCommit(newBatch, attempt + 1);
           } else {
             batch.failRemainingOperations(results);
-            batch.markComplete();
           }
-        } else {
-          batch.markComplete();
         }
         return ApiFutures.immediateFuture(null);
       }
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
@@ -488,6 +488,7 @@ public class BulkWriter {
       writeFutures.addAll(batch.getPendingFutures());
     }
     sendReadyBatches();
+    // TODO: update to use `successfulAsList()` in order to handle exceptions.
     ApiFutures.allAsList(writeFutures)
         .addListener(
             new Runnable() {
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
@@ -80,7 +80,7 @@ public class BulkWriter {
   /** The maximum number of writes that can be in a single batch. */
   public static final int MAX_BATCH_SIZE = 500;
 
-  private static final int MAX_RETRY_ATTEMPTS = 10;
+  public static final int MAX_RETRY_ATTEMPTS = 10;
 
   /**
    * The starting maximum number of operations per second as allowed by the 500/50/5 rule.
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
@@ -18,7 +18,6 @@ package com.google.cloud.firestore;
 
 import com.google.api.core.ApiAsyncFunction;
 import com.google.api.core.ApiFuture;
-import com.google.api.core.ApiFutureCallback;
 import com.google.api.core.ApiFutures;
 import com.google.api.core.CurrentMillisClock;
 import com.google.api.core.SettableApiFuture;
@@ -33,7 +32,6 @@ import com.google.common.util.concurrent.MoreExecutors;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -46,6 +44,26 @@ class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
   BulkCommitBatch(FirestoreImpl firestore, int maxBatchSize) {
     super(firestore, maxBatchSize);
   }
+
+  //  BulkCommitBatch(
+  //      FirestoreImpl firestore,
+  //      BulkCommitBatch retryBatch,
+  //      final Set<DocumentReference> docsToRetry) {
+  //    super(firestore);
+  //    this.writes.addAll(
+  //        FluentIterable.from(retryBatch.writes)
+  //            .filter(
+  //                new Predicate<WriteOperation>() {
+  //                  @Override
+  //                  public boolean apply(WriteOperation writeOperation) {
+  //                    return docsToRetry.contains(writeOperation.documentReference);
+  //                  }
+  //                })
+  //            .toList());
+  //
+  //    this.state = BatchState.READY_TO_SEND;
+  //    // TODO: figure out how to pipe completeFuture through into the new batch
+  //  }
 
   ApiFuture<WriteResult> wrapResult(ApiFuture<WriteResult> result) {
     return result;
@@ -104,7 +122,7 @@ public class BulkWriter {
 
   private final FirestoreImpl firestore;
 
-//  private final ScheduledExecutorService firestoreExecutor;
+  //  private final ScheduledExecutorService firestoreExecutor;
 
   private final ExponentialRetryAlgorithm backoff;
   private TimedAttemptSettings nextAttempt;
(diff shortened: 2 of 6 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,28 +583,84 @@ public void run() {
...
-              sendReadyBatches();
-            }
-          },
-          MoreExecutors.directExecutor());
-    } catch (Exception e) {
-      batch.processResults(new ArrayList<BatchWriteResult>(), e);
+            });
+  }
+
+  private void bulkCommit(BulkCommitBatch batch) {
```
thebrianchen, line 600: What do you think about having the retry logic live in `BulkWriter` instead of inside `UpdateBuilder`? I originally wanted to take care of everything in `UpdateBuilder`, but then I realized that doing so would prohibit creating a new `BulkCommitBatch` via constructor.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,28 +583,84 @@ public void run() {
...
+                        // Map the failure to each individual write's result.
+                        for (String path : finalBatch.getPendingDocs()) {
+                          if (e instanceof FirestoreException) {
+                            results.add(
+                                new BatchWriteResult(
+                                    path,
+                                    null,
+                                    ((FirestoreException) e).getStatus(),
+                                    e.getMessage()));
+                          } else {
```
thebrianchen, line 624: I wasn't sure how to handle the case where a non-FirestoreException was thrown. I added an additional field for an Exception in the event that the Exception thrown does not have a `Status`, but I wonder if this could be cleaner.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
...
-    return resultsMap.size();
+  int getPendingOperationCount() {
+    return pendingOperations.size();
+  }
+
+  /**
+   * Removes all operations not specified in documentsToRetry from the batch. Marks the batch as
+   * READY_TO_SEND in order to allow the next batch to be sent.
+   */
+  void sliceBatchForRetry(final Set<String> documentsToRetry) {
```
thebrianchen, line 758: Oh man. I really tried to use the BulkCommitBatch constructor directly to properly imitate Node. However, since the `BulkCommitBatch` logic lives inside `UpdateBuilder`, it got pretty ugly. I had to add a bunch of getters, and managing the state on a `BulkCommitBatch`  becomes a tricky problem / potential for future bugs. I think slicing the batch here is the lesser of the two evils, and I'm open to separating out the batch state override.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
...
+          .setException(FirestoreException.serverRejected(result.getStatus(), result.getMessage()));
+      this.pendingOperations.remove(result.getKey());
+    }
+  }
+
+  private boolean shouldRetry(@Nullable Status status) {
+    if (status == null) {
+      return false;
+    }
+    Set<Code> codes = FirestoreSettings.newBuilder().batchWriteSettings().getRetryableCodes();
```
thebrianchen, line 813: Is this part clear, or should I add some code comments?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,28 +583,84 @@ public void run() {
...
+                        }
+                        return results;
+                      }
+                    }
+                  },
+                  nextAttempt.getRandomizedRetryDelay().toMillis(),
+                  TimeUnit.MILLISECONDS);
+
+      try {
+        results = attemptBulkCommit.get();
```
schmidt-sebastian, line 636: I think this code needs to be rewritten to be a recursive function call, which would allow you to implement retry without blocking. This `.get()` blocks the thread that invokes the bulk commit, which we shouldn't do since we use the returned Task to report progress.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,28 +583,84 @@ public void run() {
...
+  }
+
+  private void bulkCommit(BulkCommitBatch batch) {
+    List<BatchWriteResult> results = new ArrayList<>();
+    for (int attempt = 0; attempt < MAX_RETRY_ATTEMPTS; ++attempt) {
+      final BulkCommitBatch finalBatch = batch;
+      ScheduledFuture<List<BatchWriteResult>> attemptBulkCommit =
+          firestore
+              .getClient()
+              .getExecutor()
```
schmidt-sebastian, line 607: Can we store this in an instance variable? I am trying to make this code look as pretty as the Node code, but it might be a lost cause until we migrate to Java 8.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,28 +583,84 @@ public void run() {
...
+              .schedule(
+                  new Callable<List<BatchWriteResult>>() {
+                    public List<BatchWriteResult> call() {
+                      List<BatchWriteResult> results = new ArrayList<>();
+                      try {
+                        return finalBatch.bulkCommit().get();
+                      } catch (Exception e) {
+                        // Map the failure to each individual write's result.
+                        for (String path : finalBatch.getPendingDocs()) {
+                          if (e instanceof FirestoreException) {
```
schmidt-sebastian, line 617: You might be able to use `ApiFutures.catching` to simplify this code.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
schmidt-sebastian: The code in `bulkCommit` should wrap all exceptions in FirestoreException.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
```diff
@@ -25,26 +25,53 @@
...
     this.message = message;
+    this.exception = null;
+  }
+
+  BatchWriteResult(
+      String key,
+      @Nullable Timestamp timestamp,
+      Status status,
+      String message,
+      Exception exception) {
```
schmidt-sebastian, line 47: We can probably drop `status`, `message` and just take a FirestoreException if we update bulkCommit to only ever throw FirestoreExceptions.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,28 +583,84 @@ public void run() {
...
+            String.format(
+                "Current batch failed at retry #%d. Num failures: %d",
+                attempt, batch.getPendingOperationCount()));
+        batch.sliceBatchForRetry(batch.getPendingDocs());
+        batch.markReadyToSend();
+
+      } else {
+        batch.markComplete();
+        return;
+      }
```
schmidt-sebastian, line 657:
```suggestion
      } else {
        break;
      }
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -51,16 +53,24 @@
  * WriteBatch}.
  */
 public abstract class UpdateBuilder<T> {
+  static class WriteOperation {
+    Write.Builder write;
+    String path;
+
+    WriteOperation(String path, Write.Builder write) {
```
schmidt-sebastian, line 60: In Java, you can use DocumentReferences since they implement `equals()`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -656,6 +666,9 @@ private T performDelete(
   /**
    * Commits all pending operations to the database and verifies all preconditions.
    *
+   * <p>If any of the writes in the batch fail with a retryable error, this method will retry the
```
schmidt-sebastian, line 669: That doesn't seem true anymore.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
     return state;
   }
 
-  Set<DocumentReference> getDocuments() {
-    return documents;
+  boolean hasPath(String path) {
```
schmidt-sebastian, line 738: This should also take a DocumentReference and likely needs to be renamed accordingly.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
...
+    return this.pendingOperations.containsKey(path);
+  }
+
+  Set<String> getPendingDocs() {
+    return this.pendingOperations.keySet();
```
schmidt-sebastian, line 743: See above.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
...
 
-    if (getOperationCount() == maxBatchSize) {
+    if (getPendingOperationCount() == maxBatchSize) {
       state = BatchState.READY_TO_SEND;
     }
 
     return result;
   }
   /** Resolves the individual operations in the batch with the results. */
   void processResults(List<BatchWriteResult> results, @Nullable Exception error) {
```
schmidt-sebastian, line 786: `error` is unused.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
schmidt-sebastian: Seems clear.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
...
+      Preconditions.checkState(
+          result.getStatus() != Status.OK, "Should not fail successful operation");
+      this.pendingOperations
+          .get(result.getKey())
+          .setException(FirestoreException.serverRejected(result.getStatus(), result.getMessage()));
+      this.pendingOperations.remove(result.getKey());
+    }
+  }
+
+  private boolean shouldRetry(@Nullable Status status) {
```
schmidt-sebastian, line 809: Ideally, `status` should not be nullable.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
...
   void processResults(List<BatchWriteResult> results, @Nullable Exception error) {
-    for (int i = 0; i < resultsMap.size(); ++i) {
-      if (error == null) {
-        SettableApiFuture<WriteResult> future = resultsMap.get(i);
-        convertBatchWriteResult(results.get(i), future);
-      } else {
-        resultsMap.get(i).setException(error);
+    for (BatchWriteResult result : results) {
+      if (result.getException() != null) {
+        pendingOperations.get(result.getKey()).setException(result.getException());
```
schmidt-sebastian, line 789: Is this also `convertBatchWriteResult`?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
...
+        pendingOperations.remove(result.getKey());
+      } else if (result.getStatus() == Status.OK || !shouldRetry(result.getStatus())) {
```
schmidt-sebastian, line 791: We should really clean up the BatchWriteStatus to only offer one way to specify success or exceptions. This code is a bit confusing as exceptions are handled in two different places based on either the exception or the status code.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -67,6 +69,16 @@
...
 
+  /** Executor that executes delayed tasks without delay. */
+  private final ScheduledExecutorService immediateExecutor =
+      new ScheduledThreadPoolExecutor(1) {
+        @Override
+        @Nonnull
+        public <V> ScheduledFuture<V> schedule(Callable<V> command, long delay, TimeUnit unit) {
+          return super.schedule(command, 0, TimeUnit.MILLISECONDS);
+        }
+      };
```
schmidt-sebastian, line 80: Nice :)

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -401,6 +418,52 @@ public void sendBatchesWhenSizeLimitIsReached() throws Exception {
...
+        new ResponseStubber() {
+          {
+            put(
+                batchWrite(
+                    set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc1"),
+                    set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc2"),
+                    set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc3")),
+                mergeResponses(
+                    failedResponse(),
+                    failedResponse(Code.UNAVAILABLE_VALUE),
```
schmidt-sebastian, line 433: Maybe update the test name to reflect that we also retry on UNAVAILABLE?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -550,15 +638,59 @@ public void individualWritesErrorIfBulkCommitFails() throws Exception {
...
+    ApiFuture<WriteResult> result = bulkWriter.set(doc1, LocalFirestoreHelper.SINGLE_FIELD_MAP);
+    bulkWriter.close().get();
+
+    try {
+      result.get();
+    } catch (Exception e) {
+      assertTrue(e.getMessage().contains("Mock batchWrite failed in test"));
+      ++opCount;
+    }
+    assertEquals(1, opCount);
```
schmidt-sebastian, line 694:
```suggestion
     Assert.fail(...);
    } catch (Exception e) {
      assertTrue(e.getMessage().contains("Mock batchWrite failed in test"));
    }
```

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -550,15 +638,59 @@ public void individualWritesErrorIfBulkCommitFails() throws Exception {
...
+
+    ApiFuture<WriteResult> result = bulkWriter.set(doc1, LocalFirestoreHelper.SINGLE_FIELD_MAP);
+    bulkWriter.close().get();
+
+    assertEquals(Timestamp.ofTimeSecondsAndNanos(3, 0), result.get().getUpdateTime());
+  }
+
+  @Test
+  public void failsWritesAfterAllRetryAttemptsFail() throws Exception {
+    doThrow(
```
schmidt-sebastian, line 677: Does the mocking layer allow us to specify that we want this function to be called 10 times?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,40 +735,88 @@ BatchState getState() {
...
+    }
+  }
+
+  private boolean shouldRetry(@Nullable Status status) {
+    if (status == null) {
+      return false;
+    }
+    Set<Code> codes = FirestoreSettings.newBuilder().batchWriteSettings().getRetryableCodes();
+    for (Code code : codes) {
+      if (code.toString().equals(status.getCode().toString())) {
```
schmidt-sebastian, line 815:
```suggestion
      if (code.equals(Code.valueOf(status.getCode().name()))) {
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
schmidt-sebastian: How about https://gist.github.com/schmidt-sebastian/48ed97084b74bf7e3dd5936eddbfc001, or is this only addressing half of the problem?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
schmidt-sebastian: This matches Node so I am fully supportive.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
schmidt-sebastian: One more thing: What happens if `firestore.getClient().getExecutor()` is multi-threaded?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
thebrianchen: removed.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
thebrianchen: I ran into two main issues: first, my currently implementation requires the batch in `bulkCommit()` to be marked `final`, which means it can't be changed directly, though I guess we could put it in an array. The second is that the `completeFuture` used to track `awaitBulkCommit()` will never resolve, unless we pipe it through somehow (is that possible?).

thebrianchen: Update: Ended up updating to use BulkCommitBatch as constructor instead of slicing, pending `successfulAsList`

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,28 +583,84 @@ public void run() {
...
-            public void run() {
-              try {
-                batch.processResults(commitFuture.get(), null);
-              } catch (Exception e) {
-                batch.processResults(new ArrayList<BatchWriteResult>(), e);
+    MoreExecutors.directExecutor()
+        .execute(
+            new Runnable() {
+              public void run() {
+                bulkCommit(batch);
```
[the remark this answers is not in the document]
thebrianchen: Update: N/A now that we are chaining with ApiFutures.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: rewritten to use ApiFuture.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
thebrianchen: Changed to use doAnswer.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,28 +605,98 @@ public void run() {
...
+          if (attempt < MAX_RETRY_ATTEMPTS) {
+            nextAttempt = backoff.createNextAttempt(nextAttempt);
+            BulkCommitBatch newBatch =
+                new BulkCommitBatch(firestore, batch, batch.getPendingDocuments());
+            return bulkCommit(newBatch, attempt + 1);
+          } else {
+            batch.failRemainingOperations(results);
+          }
+        }
+        return ApiFutures.immediateFuture(null);
```
schmidt-sebastian, line 660: This confused me during the review. There is a strange interaction model going on here.

`batch.processResults` changes the internal state, which reduces the number of pending operations (this is not obvious from the callsite). `new BulkCommitBatch` then uses this filtered list of pending operations, but also uses the unfiltered list of writes. 

While this matches what Node does, a lot of the back and forth in Node happens in the same codepath, and here these interactions all cross into and out of UpdateBuilder.

One improvement suggestion:


```suggestion
         batch.processResults(results);
         List<DocumentReference> pendingOps = batch.getRemainingOperations();
        if (!pendingOps.isEmpty()) {
          logger.log(
              Level.WARNING,
              String.format(
                  "Current batch failed at retry #%d. Num failures: %d",
                  attempt,pendingOps.size())));

          if (attempt < MAX_RETRY_ATTEMPTS) {
            nextAttempt = backoff.createNextAttempt(nextAttempt);
            BulkCommitBatch newBatch =
                new BulkCommitBatch(firestore, batch, pendingOps);
            return bulkCommit(newBatch, attempt + 1);
          } else {
            batch.failRemainingOperations(results);
          }
        }
        return ApiFutures.immediateFuture(null);
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -41,6 +47,26 @@
...
+            .filter(
+                new Predicate<WriteOperation>() {
+                  @Override
+                  public boolean apply(WriteOperation writeOperation) {
+                    return docsToRetry.contains(writeOperation.documentReference);
+                  }
+                })
+            .toList());
+
+    this.state = retryBatch.state;
```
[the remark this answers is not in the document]
schmidt-sebastian: Hm. The new BulkCommitBatch hasn't been sent though, and hence the state is a bit misleading. Would it break anything if we used READY_TO_SEND here?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -85,8 +117,11 @@ boolean allowDuplicateDocs() {
   /** The maximum number of writes that can be in a single batch. */
   private int maxBatchSize = MAX_BATCH_SIZE;
 
-  /** A queue of batches to be written. */
-  private final List<BulkCommitBatch> batchQueue = new ArrayList<>();
+  /**
+   * A queue of batches to be written. Use a synchronized list to avoid multi-thread concurrent
+   * modification errors.
+   */
```
schmidt-sebastian, line 123:
```suggestion
  /**
   * A queue of batches to be written. Use a synchronized list to avoid multi-thread concurrent
   * modification errors (as this list is modified from both the user thread and the network thread).
   */
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -568,27 +612,109 @@ public void run() {
...
+
+    public ProcessBulkCommitCallback(BulkCommitBatch batch, int attempt) {
+      this.batch = batch;
+      this.attempt = attempt;
+    }
+
+    @Override
+    public ApiFuture<Void> apply(List<BatchWriteResult> results) {
+      batch.processResults(results);
+      Set<DocumentReference> pendingOps = batch.getPendingDocuments();
```
schmidt-sebastian, line 702: The logic is much cleaner now, but still not self-explanatory. I think this can be solved by naming though:

```
batch.processResults(results);
Set<DocumentReference> remainingOp = batch.getRemainingOperations();
```

Or:
```
batch.processResultsAndRemoveSuccesfulResultsFromPendingOperationsBernie2016(results);
Set<DocumentReference> remainingOp = batch. getPendingDocuments();
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -698,9 +701,17 @@ private T performDelete(
...
+              @Nullable Exception exception;
+              if (code == Status.OK) {
+                updateTime = Timestamp.fromProto(writeResult.getUpdateTime());
+                exception = null;
+              } else {
+                updateTime = null;
+                exception = FirestoreException.serverRejected(code, status.getMessage());
+              }
+              result.add(
+                  new BatchWriteResult(writes.get(i).documentReference, updateTime, exception));
```
schmidt-sebastian, line 714:
```suggestion
              Timestamp updateTime = null;
              Exception exception = null;
              if (Status.OK.equals(code)) {
                updateTime = Timestamp.fromProto(writeResult.getUpdateTime());
              } else {
                exception = FirestoreException.serverRejected(code, status.getMessage());
              }
              result.add(
                  new BatchWriteResult(writes.get(i).documentReference, updateTime, exception));
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -723,58 +734,78 @@ BatchState getState() {
...
   }
-  /** Resolves the individual operations in the batch with the results. */
-  void processResults(List<BatchWriteResult> results, @Nullable Exception error) {
-    for (int i = 0; i < resultsMap.size(); ++i) {
-      if (error == null) {
-        SettableApiFuture<WriteResult> future = resultsMap.get(i);
-        convertBatchWriteResult(results.get(i), future);
-      } else {
-        resultsMap.get(i).setException(error);
+  /**
```
schmidt-sebastian, line 767: Missing empty line.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -392,13 +410,84 @@ public void sendBatchesWhenSizeLimitIsReached() throws Exception {
...
+    }
     assertEquals(Timestamp.ofTimeSecondsAndNanos(2, 0), result2.get().getUpdateTime());
     assertEquals(Timestamp.ofTimeSecondsAndNanos(3, 0), result3.get().getUpdateTime());
+
+    List<BatchWriteRequest> requests = batchWriteCapture.getAllValues();
+    assertEquals(responseStubber.size(), requests.size());
+  }
+
+  @Test
+  public void allWritesCompleteWhenFlushCompletes() throws Exception {
```
schmidt-sebastian, line 469: Is this more "flushCompletesWhenallWritesComplete"?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -550,15 +634,67 @@ public void individualWritesErrorIfBulkCommitFails() throws Exception {
...
+    bulkWriter.close().get();
+
+    try {
+      result.get();
+      Assert.fail("Expected set() operation to fail");
+    } catch (Exception e) {
+      assertTrue(e.getMessage().contains("Mock batchWrite failed in test"));
+      assertEquals(retryAttempts[0], BulkWriter.MAX_RETRY_ATTEMPTS + 1);
+    }
+  }
```
schmidt-sebastian, line 699: Can we add a test that:

- Adds two writes
- Calls flush
- Fails one write with retryable error, let's the other succeed
- Retries write which succeeds
- Test verifies that flush() only finishes now

I am not 100% convinced that this works right now.

thebrianchen: Added test.

PR #293, opened 2020-07-11
feat: Implementation of Firestore Bundle Builder

wu-hui:


google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1057,6 +1058,71 @@ public Query endAt(@Nonnull DocumentSnapshot snapshot) {
...
+                .build();
+        structuredQuery.setStartAt(cursor);
+      }
+    }
+
+    return structuredQuery;
+  }
+
+  /**
+   * Builds a query to saved in a bundle file.
```
BenWhitehead, line 1108:
```suggestion
   * Builds a {@link BundledQuery} that is able to be saved in a bundle file.
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1057,6 +1058,71 @@ public Query endAt(@Nonnull DocumentSnapshot snapshot) {
...
+   *
+   * <p>This will not do client translation (limitToLast order flip, for example),
```
BenWhitehead, line 1110:
```suggestion
   * <p>This will not perform any limitToLast order flip, as {@link BundledQuery} has first class
   * representation via {@link BundledQuery.LimitType}
```

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1426,4 +1438,80 @@ public void onCompleted() {
...
+        BundledDocumentMetadata.newBuilder()
+            .setExists(false)
+            .setName(fullPath(randomDoc, firestore.getOptions()))
+            .setReadTime(snapshot.getReadTime().toProto())
+            .build(),
+        elements.get(1).getDocumentMetadata());
+  }
+
+  @Test
+  public void testBuildingBundleWithLimitAndLimitToLastQueries() throws Exception {
```
BenWhitehead, line 1467: Can we split this test into two, one for the limit First, and one for limit last? If either case ever fails it'd be nice to tell independently which is failing.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -0,0 +1,228 @@
...
+import com.google.firestore.proto.NamedQuery;
+import com.google.firestore.v1.Document;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
```
BenWhitehead, line 34: This library currently still supports java 7, so we can't use `java.util.Optional` here. It looks like it is only needed for page-private or lower reasons, so can probably instead use `com.google.common.base.Optional` from Guava.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -0,0 +1,228 @@
...
+  /** Returns the bundle content as a readonly {@link ByteBuffer}. */
+  public ByteBuffer toByteBuffer() {
+    return ByteBuffer.wrap(bundleData).asReadOnlyBuffer();
+  }
+}
+
+/**
+ * Convenient class to hold both the metadata and the actual content of a document to be bundled.
+ */
+class BundledDocument {
```
BenWhitehead, line 208: It is very uncommon to have more than one top level class per java file. Can we make it a static inner class of `FirestoreBundle`, or move this class to its own file?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -0,0 +1,228 @@
...
+  private BundledDocumentMetadata metadata;
+  private Document document;
```
BenWhitehead, line 210: This class appears to be write-once, can we mark these fields as `final`?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
```diff
@@ -0,0 +1,325 @@
...
+import static com.google.cloud.firestore.LocalFirestoreHelper.DOCUMENT_NAME;
+import static com.google.cloud.firestore.LocalFirestoreHelper.SINGLE_FIELD_SNAPSHOT;
+import static com.google.cloud.firestore.LocalFirestoreHelper.UPDATED_SINGLE_FIELD_SNAPSHOT;
+import static com.google.cloud.firestore.LocalFirestoreHelper.bundleToElementList;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import com.google.cloud.firestore.spi.v1.FirestoreRpc;
+import com.google.common.collect.Lists;
```
BenWhitehead, line 29: Same as in the prod file.
```suggestion
import com.google.common.collect.Lists;
```

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
```diff
@@ -0,0 +1,325 @@
...
+        // because it is the largest read time came across.
+        UPDATED_SINGLE_FIELD_SNAPSHOT.getReadTime().toProto(),
+        /*totalDocuments*/ 1,
+        /*expectEmptyContent*/ false);
+
+    verifyDocumentAndMeta(
+        elements.get(1).getDocumentMetadata(),
+        elements.get(2).getDocument(),
+        DOCUMENT_NAME,
+        Lists.newArrayList(),
```
BenWhitehead, line 192: The expected type parameter isn't carried through here (the type of the list is derived from the elements passed to the method), we have to specify it so the compiler is happy.
```suggestion
        Lists.<String>newArrayList(),
```

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
```diff
@@ -0,0 +1,325 @@
...
+        elements.get(0).getMetadata(),
+        SINGLE_FIELD_SNAPSHOT.getReadTime().toProto(),
+        /*totalDocuments*/ 1,
+        /*expectEmptyContent*/ false);
+
+    verifyDocumentAndMeta(
+        elements.get(1).getDocumentMetadata(),
+        elements.get(2).getDocument(),
+        DOCUMENT_NAME,
+        Lists.newArrayList(),
```
BenWhitehead, line 287: Same as above
```suggestion
        Lists.<String>newArrayList(),
```

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
```diff
@@ -0,0 +1,325 @@
...
+    verifyMetadata(
+        elements.get(0).getMetadata(),
+        UPDATED_SINGLE_FIELD_SNAPSHOT.getReadTime().toProto(),
+        /*totalDocuments*/ 1,
+        /*expectEmptyContent*/ false);
+    verifyDocumentAndMeta(
+        elements.get(1).getDocumentMetadata(),
+        elements.get(2).getDocument(),
+        DOCUMENT_NAME,
+        Lists.newArrayList(),
```
BenWhitehead, line 306: Same as above
```suggestion
        Lists.<String>newArrayList(),
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -0,0 +1,228 @@
...
+    public Builder add(DocumentSnapshot documentSnapshot) {
+      return add(documentSnapshot, Optional.empty());
+    }
+
+    private Builder add(DocumentSnapshot documentSnapshot, Optional<String> queryName) {
+      String documentName = documentSnapshot.getReference().getName();
+      BundledDocument originalDocument = documents.get(documentSnapshot.getReference().getName());
+      List<String> queries =
+          originalDocument == null
+              ? Lists.newArrayList()
```
BenWhitehead, line 77:
```suggestion
              ? Lists.<String>newArrayList()
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -0,0 +1,228 @@
...
+              : Lists.newArrayList(originalDocument.getMetadata().getQueriesList());
+
+      // Update with document built from `documentSnapshot` because it is newer.
+      if (originalDocument == null
+          || documentSnapshot
+                  .getReadTime()
+                  .compareTo(Timestamp.fromProto(originalDocument.getMetadata().getReadTime()))
```
BenWhitehead, line 84: This call to compare to should guard against `getReadTime()` returning null, since there doesn't appear to be a previous check ensuring it's non-null.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -0,0 +1,228 @@
...
+    // Id of the bundle.
+    private String id;
+    // Resulting documents for the bundle, keyed by full document path.
+    private Map<String, BundledDocument> documents = new HashMap<>();
+    // Named queries saved in the bundle, keyed by query name.
+    private Map<String, NamedQuery> namedQueries = new HashMap<>();
+    // The latest read time among all bundled documents and queries.
+    private Timestamp latestReadTime = Timestamp.MIN_VALUE;
+    // Printer to encode protobuf objects into JSON string.
+    private JsonFormat.Printer printer = JsonFormat.printer();
```
BenWhitehead, line 55: If JsonFormat.Printer is thread safe (a quick glance though it seems like it is, but I don't know for sure), this field could be turned into a constant and have a single instance shared across all builders.

BenWhitehead (changes requested):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
wu-hui: I gave it a try, and it looks awkward. I am inclining to not check it here, because the only place setting `originalDocument` is below, and the need to defend against that is not very strong IMHO.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
wu-hui: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
wu-hui: Done, metadata is set above though.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
wu-hui: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
wu-hui: Done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
wu-hui: Is this a typo?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
wu-hui: Done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
wu-hui: Done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
wu-hui: Done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
wu-hui: My thought was this provides a good contrast and helps reader understand two cases.

I don't feel strong, I can split if you insist.

wu-hui (commented):
Thanks for the review!

It is now updated with bundle.proto from googleapis/third_party as well.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
BenWhitehead: Yeah, ignore this.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
BenWhitehead: If the code overhead is minimal I'd prefer the two separate tests for clarity of individual scenarios.

google-cloud-firestore/pom.xml
```diff
@@ -157,6 +157,12 @@
       <version>3.11</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>proto-google-cloud-firestore-bundle-v1</artifactId>
+      <version>2.1.1-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
BenWhitehead, line 165: Due to some overarching complexities in how our poms get built, bundled and released with the broader GCP libraries this declaration needs to change a little bit. I'll fix up all the pom files and push a commit to this branch with all the necessary changes, we can then review if you're interested in the details.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -0,0 +1,228 @@
...
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Represents a Firestore data bundle with results from the given document and query snapshots. */
+public final class FirestoreBundle {
+
+  static int BUNDLE_SCHEMA_VERSION = 1;
```
BenWhitehead, line 39:
```suggestion
  static final int BUNDLE_SCHEMA_VERSION = 1;
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -0,0 +1,228 @@
...
+  // Printer to encode protobuf objects into JSON string.
+  private static JsonFormat.Printer PRINTER = JsonFormat.printer();
```
BenWhitehead, line 41:
```suggestion
  private static final JsonFormat.Printer PRINTER = JsonFormat.printer();
```

Constants should always be final so they can't be modified.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
BenWhitehead: The `compareTo` method on Timestamp expects a non-null argument (and doesn't assert it unfortunately) and there is the possibility that `getReadTime()` returns a null value, so we do in fact need to guard this check so that it doesn't result in an exception.

A similar scenario is present below on line 88 (in this commit) where it's possible a null value is returned before calling `toProto()`.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/FirestoreBundleTest.java
BenWhitehead: Yes please ignore.

BenWhitehead (changes requested):
A couple small things, and the remaining comment about `getReadTime` possibly causing an exception.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
wu-hui: Done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
wu-hui: Done.

BenWhitehead (approved):

PR #296, opened 2020-07-15
fix: use test credentials when connecting to the Emulator from Admin SDK

schmidt-sebastian:
Fixes https://github.com/firebase/firebase-admin-java/issues/452


schmidt-sebastian:
Unfortunately, we don't have any automated regressions tests that involve multiple projects. I verified with the customer's repo that their problem no longer exists and verified that our integration tests still pass against the Emulator even when the Admin SDK is not involved (well, there are two failures, but they existed prior to this change).

PR #320, opened 2020-08-06
feat: add support for read-only transactions in TransactionOptions

BenWhitehead:
* Add new typesafe builders for read-only `TransactionOptions.readOnlyOptionsBuilder` and read-write `TransactionOptions.readWriteOptionsBuilder` transactions in TransactionOptions
  * These new builders ensure only those parameters relevant to the respective type of transaction are available
* Deprecate existing `TransactionOptions.create(...)` methods in favor of the new builders
* Deprecate existing and annotate @InternalApi `TransactionOptions.getNumberOfAttempts`
* Update Transaction and TransactionRunner to use TransactionOptions directly

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
@@ -63,7 +63,7 @@ public final class Transaction extends UpdateBuilder<Transaction> {
     ApiFuture<T> updateCallback(Transaction transaction);
   }
 
-  private final TransactionOptions originalTransactionOptions;
+  private final TransactionOptions transactionOptions;
   @Nullable private final ByteString previousTransactionId;
   private ByteString transactionId;
 
@@ -72,7 +72,7 @@ public final class Transaction extends UpdateBuilder<Transaction> {
       TransactionOptions transactionOptions,
       @Nullable Transaction previousTransaction) {
     super(firestore);
-    this.originalTransactionOptions = transactionOptions;
+    this.transactionOptions = transactionOptions;
     this.previousTransactionId =
         previousTransaction != null ? previousTransaction.transactionId : null;
   }
@@ -86,14 +86,14 @@ public final class Transaction extends UpdateBuilder<Transaction> {
     BeginTransactionRequest.Builder beginTransaction = BeginTransactionRequest.newBuilder();
     beginTransaction.setDatabase(firestore.getDatabaseName());
 
-    if (originalTransactionOptions.getType() == TransactionOptionsType.READ_WRITE
+    if (transactionOptions.getType() == TransactionOptionsType.READ_WRITE
         && previousTransactionId != null) {
       beginTransaction
           .getOptionsBuilder()
           .getReadWriteBuilder()
           .setRetryTransaction(previousTransactionId);
-    } else if (originalTransactionOptions.getType() == TransactionOptionsType.READ_ONLY) {
-      final ReadOnly.Builder builder = originalTransactionOptions.getReadOnly().toProtoBuilder();
+    } else if (transactionOptions.getType() == TransactionOptionsType.READ_ONLY) {
+      final ReadOnly.Builder builder = transactionOptions.getReadOnly().toProtoBuilder();
       beginTransaction.getOptionsBuilder().setReadOnly(builder);
     }
 
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
@@ -25,6 +25,7 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.google.firestore.v1.BeginTransactionRequest;
 import com.google.firestore.v1.BeginTransactionResponse;
 import com.google.firestore.v1.RollbackRequest;
+import com.google.firestore.v1.TransactionOptions.ReadOnly;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Empty;
 import java.util.List;
@@ -91,12 +92,12 @@ public final class Transaction extends UpdateBuilder<Transaction> {
           .getOptionsBuilder()
           .getReadWriteBuilder()
           .setRetryTransaction(previousTransactionId);
-    } else if (TransactionOptionsType.READ_ONLY.equals(transactionOptions.getType())
-        && transactionOptions.getReadTime() != null) {
-      beginTransaction
-          .getOptionsBuilder()
-          .getReadOnlyBuilder()
-          .setReadTime(transactionOptions.getReadTime());
+    } else if (TransactionOptionsType.READ_ONLY.equals(transactionOptions.getType())) {
+      final ReadOnly.Builder readOnlyBuilder = ReadOnly.newBuilder();
+      if (transactionOptions.getReadTime() != null) {
+        readOnlyBuilder.setReadTime(transactionOptions.getReadTime());
+      }
+      beginTransaction.getOptionsBuilder().setReadOnly(readOnlyBuilder);
     }
 
     ApiFuture<BeginTransactionResponse> transactionBeginFuture =
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
@@ -25,7 +25,6 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.google.firestore.v1.BeginTransactionRequest;
 import com.google.firestore.v1.BeginTransactionResponse;
 import com.google.firestore.v1.RollbackRequest;
-import com.google.firestore.v1.TransactionOptions.ReadOnly;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Empty;
 import java.util.List;
@@ -92,9 +91,8 @@ public final class Transaction extends UpdateBuilder<Transaction> {
           .getOptionsBuilder()
           .getReadWriteBuilder()
           .setRetryTransaction(previousTransactionId);
-    } else if (TransactionOptionsType.READ_ONLY.equals(transactionOptions.getType())) {
-      final ReadOnly builder = transactionOptions.getReadOnly().toProto();
-      beginTransaction.getOptionsBuilder().setReadOnly(builder);
+    } else if (TransactionOptionsType.READ_ONLY.equals(transactionOptions.getType()) && transactionOptions.getReadTime()!= null) {
+        beginTransaction.getOptionsBuilder().getReadOnlyBuilder().setReadTime(transactionOptions.getReadTime());
     }
 
     ApiFuture<BeginTransactionResponse> transactionBeginFuture =
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
@@ -86,13 +86,13 @@ public final class Transaction extends UpdateBuilder<Transaction> {
     BeginTransactionRequest.Builder beginTransaction = BeginTransactionRequest.newBuilder();
     beginTransaction.setDatabase(firestore.getDatabaseName());
 
-    if (transactionOptions.getType() == TransactionOptionsType.READ_WRITE
+    if (TransactionOptionsType.READ_WRITE.equals(transactionOptions.getType())
         && previousTransactionId != null) {
       beginTransaction
           .getOptionsBuilder()
           .getReadWriteBuilder()
           .setRetryTransaction(previousTransactionId);
-    } else if (transactionOptions.getType() == TransactionOptionsType.READ_ONLY) {
+    } else if (TransactionOptionsType.READ_ONLY.equals(transactionOptions.getType())) {
       final ReadOnly.Builder builder = transactionOptions.getReadOnly().toProtoBuilder();
       beginTransaction.getOptionsBuilder().setReadOnly(builder);
     }
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
@@ -19,11 +19,14 @@ package com.google.cloud.firestore;
 import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
+import com.google.cloud.firestore.TransactionOptions.EitherReadOnlyOrReadWrite;
+import com.google.cloud.firestore.TransactionOptions.TransactionOptionsType;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.firestore.v1.BeginTransactionRequest;
 import com.google.firestore.v1.BeginTransactionResponse;
 import com.google.firestore.v1.RollbackRequest;
+import com.google.firestore.v1.TransactionOptions.ReadOnly;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Empty;
 import java.util.List;
@@ -61,12 +64,18 @@ public final class Transaction extends UpdateBuilder<Transaction> {
     ApiFuture<T> updateCallback(Transaction transaction);
   }
 
+  private final TransactionOptions originalTransactionOptions;
+  @Nullable private final ByteString previousTransactionId;
   private ByteString transactionId;
-  private @Nullable ByteString previousTransactionId;
 
-  Transaction(FirestoreImpl firestore, @Nullable Transaction previousTransaction) {
+  Transaction(
+      FirestoreImpl firestore,
+      TransactionOptions transactionOptions,
+      @Nullable Transaction previousTransaction) {
     super(firestore);
-    previousTransactionId = previousTransaction != null ? previousTransaction.transactionId : null;
+    this.originalTransactionOptions = transactionOptions;
+    this.previousTransactionId =
+        previousTransaction != null ? previousTransaction.transactionId : null;
   }
 
   Transaction wrapResult(ApiFuture<WriteResult> result) {
@@ -78,11 +87,15 @@ public final class Transaction extends UpdateBuilder<Transaction> {
     BeginTransactionRequest.Builder beginTransaction = BeginTransactionRequest.newBuilder();
     beginTransaction.setDatabase(firestore.getDatabaseName());
 
-    if (previousTransactionId != null) {
+    final EitherReadOnlyOrReadWrite options = originalTransactionOptions.getOptions();
+    if (options.getType() == TransactionOptionsType.READ_WRITE && previousTransactionId != null) {
       beginTransaction
           .getOptionsBuilder()
           .getReadWriteBuilder()
           .setRetryTransaction(previousTransactionId);
+    } else if (options.getType() == TransactionOptionsType.READ_ONLY) {
+      final ReadOnly.Builder builder = options.getReadOnly().toProtoBuilder();
+      beginTransaction.getOptionsBuilder().setReadOnly(builder);
     }
 
     ApiFuture<BeginTransactionResponse> transactionBeginFuture =
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
@@ -19,7 +19,6 @@ package com.google.cloud.firestore;
 import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
-import com.google.cloud.firestore.TransactionOptions.EitherReadOnlyOrReadWrite;
 import com.google.cloud.firestore.TransactionOptions.TransactionOptionsType;
 import com.google.common.base.Preconditions;
 import com.google.common.util.concurrent.MoreExecutors;
@@ -87,14 +86,13 @@ public final class Transaction extends UpdateBuilder<Transaction> {
     BeginTransactionRequest.Builder beginTransaction = BeginTransactionRequest.newBuilder();
     beginTransaction.setDatabase(firestore.getDatabaseName());
 
-    final EitherReadOnlyOrReadWrite options = originalTransactionOptions.getOptions();
-    if (options.getType() == TransactionOptionsType.READ_WRITE && previousTransactionId != null) {
+    if (originalTransactionOptions.getType() == TransactionOptionsType.READ_WRITE && previousTransactionId != null) {
       beginTransaction
           .getOptionsBuilder()
           .getReadWriteBuilder()
           .setRetryTransaction(previousTransactionId);
-    } else if (options.getType() == TransactionOptionsType.READ_ONLY) {
-      final ReadOnly.Builder builder = options.getReadOnly().toProtoBuilder();
+    } else if (originalTransactionOptions.getType() == TransactionOptionsType.READ_ONLY) {
+      final ReadOnly.Builder builder = originalTransactionOptions.getReadOnly().toProtoBuilder();
       beginTransaction.getOptionsBuilder().setReadOnly(builder);
     }
 
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
@@ -93,7 +93,7 @@ public final class Transaction extends UpdateBuilder<Transaction> {
           .getReadWriteBuilder()
           .setRetryTransaction(previousTransactionId);
     } else if (TransactionOptionsType.READ_ONLY.equals(transactionOptions.getType())) {
-      final ReadOnly.Builder builder = transactionOptions.getReadOnly().toProtoBuilder();
+      final ReadOnly builder = transactionOptions.getReadOnly().toProto();
       beginTransaction.getOptionsBuilder().setReadOnly(builder);
     }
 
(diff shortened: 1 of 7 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
```diff
@@ -61,24 +64,34 @@
...
-  private @Nullable ByteString previousTransactionId;
 
-  Transaction(FirestoreImpl firestore, @Nullable Transaction previousTransaction) {
+  Transaction(
+      FirestoreImpl firestore,
+      TransactionOptions transactionOptions,
+      @Nullable Transaction previousTransaction) {
     super(firestore);
-    previousTransactionId = previousTransaction != null ? previousTransaction.transactionId : null;
+    this.originalTransactionOptions = transactionOptions;
```
schmidt-sebastian, line 76: I am surprised this isn't just `transactionOptions`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,98 @@
...
+  private final EitherReadOnlyOrReadWrite options;
 
-  TransactionOptions(int maxAttempts, Executor executor) {
-    this.numberOfAttempts = maxAttempts;
+  TransactionOptions(Executor executor, EitherReadOnlyOrReadWrite options) {
     this.executor = executor;
+    this.options = options;
   }
 
+  /**
```
schmidt-sebastian, line 45: Read-Only transactions can fail with ABORTED too (if a read-write transaction holds a lock) and hence I think it still makes sense to retry.

Please correct me if I am wrong, since I would think that you know more about this than I do.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,98 @@
...
+   * @return int Representing the initial number of attempts a read-write transaction will be
+   *     attempted
```
schmidt-sebastian, line 47:
```suggestion
   * @return the initial number of attempts a read-write transaction will be
   *     attempted
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,98 @@
...
+   * @deprecated As of v2.0.0, only applicable to Read-Write transactions. Use {@link
```
schmidt-sebastian, line 48: I was once told that we should use lowercase here.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,98 @@
...
   public int getNumberOfAttempts() {
-    return numberOfAttempts;
+    if (options.getType() == TransactionOptionsType.READ_WRITE) {
+      return options.getReadWrite().numberOfAttempts;
+    } else {
+      return 1;
+    }
   }
 
+  /** @return Executor Executor to be used to run user callbacks on */
```
schmidt-sebastian, line 61:
```suggestion
  /** @return Executor to be used to run user callbacks on */
```

Or not?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,98 @@
...
+  @InternalApi
+  EitherReadOnlyOrReadWrite getOptions() {
+    return options;
+  }
+
   /**
-   * Create a default set of options suitable for most use cases. Transactions will be attempted 5
-   * times.
+   * Create a default set of ReadWrite options suitable for most use cases. Transactions will be
+   * attempted 5 times.
```
schmidt-sebastian, line 75:
```suggestion
   * Create a default set of options suitable for most use cases. Transactions will be opened as ReadWrite transactions and attempted up to 5 times.
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +116,205 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+    @SuppressWarnings("unchecked")
+    public B setExecutor(@Nullable Executor executor) {
+      this.executor = executor;
+      return (B) this;
+    }
+
+    @Nonnull
+    public abstract TransactionOptions build();
+
+    static ReadOnlyOptionsBuilder readOnlyBuilder() {
```
schmidt-sebastian, line 164: I am wondering if we need to carry the separation of ReadOnly versus ReadWrite this far. Could we use a single Builder that has a "mode" setting?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +116,205 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+      return new ReadOnlyOptionsBuilder(null, null);
+    }
+
+    static ReadWriteOptionsBuilder readWriteBuilder() {
+      return new ReadWriteOptionsBuilder(null, DEFAULT_NUM_ATTEMPTS);
+    }
+  }
+
+  public static final class ReadOnlyOptionsBuilder
```
schmidt-sebastian, line 173: So that I learn something, do you want to add docs for these methods and classes? Things to point out:

- ReadOnly transactions are ... well, read-only.
- Do they grab exclusive locks? I think they might not.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
```diff
@@ -58,39 +60,53 @@
...
   private Transaction transaction;
   private int attemptsRemaining;
 
   /**
    * @param firestore The active Firestore instance
    * @param userCallback The user provided transaction callback
-   * @param userCallbackExecutor The executor to run the user callback on
-   * @param numberOfAttempts The total number of attempts for this transaction
+   * @param transactionOptions The set options determining which executor the {@code userCallback}
+   *     is run on, whether the transaction is read-write or read-only
```
schmidt-sebastian, line 76:
```suggestion
   * @param transactionOptions The options determining which executor the {@code userCallback}
   *     is run on and whether the transaction is read-write or read-only
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
```diff
@@ -58,39 +60,53 @@
...
     this.userCallback = userCallback;
-    this.attemptsRemaining = numberOfAttempts;
-    this.userCallbackExecutor = userCallbackExecutor;
+    final EitherReadOnlyOrReadWrite options = transactionOptions.getOptions();
+    switch (options.getType()) {
+      case READ_WRITE:
+        this.attemptsRemaining = options.getReadWrite().getNumberOfAttempts();
+        break;
+      case READ_ONLY:
+        this.attemptsRemaining = 1;
```
schmidt-sebastian, line 93: As asked before, is this required?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +116,205 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+    @Nonnull
+    @Override
+    public TransactionOptions build() {
+      return new TransactionOptions(
+          executor, EitherReadOnlyOrReadWrite.readWrite(new ReadWriteOptions(numberOfAttempts)));
+    }
+  }
+
+  // TODO: Refactor to use Optional when java7 support is dropped
+  static final class EitherReadOnlyOrReadWrite {
```
schmidt-sebastian, line 235: Could this class be merged with TransactionOptions and provide the internal state via package-private methods?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +116,205 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+      super(executor);
+      this.readTime = readTime;
+    }
+
+    @Nullable
+    public TimestampOrBuilder getReadTime() {
+      return readTime;
+    }
+
+    public ReadOnlyOptionsBuilder setReadTime(@Nullable TimestampOrBuilder readTime) {
```
schmidt-sebastian, line 187: Al of this needs documentation, which can probably just be copied from the Proto.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1306,6 +1315,93 @@ public void deleteNestedFieldUsingFieldPath() throws Exception {
...
+
+    try {
+      runTransaction.get(10, TimeUnit.SECONDS);
+    } catch (ExecutionException e) {
+      final Throwable rootCause = ExceptionUtils.getRootCause(e);
+      assertThat(rootCause).isInstanceOf(StatusRuntimeException.class);
+      final StatusRuntimeException invalidArgument = (StatusRuntimeException) rootCause;
+      final Status status = invalidArgument.getStatus();
+      assertThat(status.getCode()).isEqualTo(Code.FAILED_PRECONDITION);
+      assertThat(status.getDescription()).contains("old");
```
schmidt-sebastian, line 1401: Did you happen to run these tests against the Emulator as well? It sometimes returns slightly different error messages.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
BenWhitehead: Based on the lack of a `retry_transaction` like [`ReadWrite`](https://github.com/googleapis/java-firestore/blob/0c841f9b8c03aaeceb2762492daeeda9f8efdd66/proto-google-cloud-firestore-v1/src/main/proto/google/firestore/v1/common.proto#L59-L62), I interpreted it as not safe to retry and instead would require the application layer to try again. Also if a `readTime` is specified it is not allowed to be more than 60 seconds ago from the time it hits the server, so if multiple retries were attempted (especially if the timeframe was smaller than 60 seconds) it's possible the specified timestamp would fallout of the valid time frame.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
BenWhitehead: Given `TransactionRunner` creates a new instance for each retry, I thought it was useful to name this field to signify it's from the first attempt. Don't feel strongly about it though if you'd prefer I change it.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
BenWhitehead: fixed

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
BenWhitehead: Updated throughout

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
BenWhitehead: Fixed

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
BenWhitehead: Incorporated

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
BenWhitehead: incorporated

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
BenWhitehead: Yup, ran them against the emulator and they all passed.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
BenWhitehead: Most likely, yes. I was writing this in such a way that it'd be able to take advantage of new java features once we drop support for 7 but it may not be worth it right now.

I'll see what I can do to pull it out.

BenWhitehead: Fixed

BenWhitehead:
I'll flush out the java docs for the builder methods next week.

If we decide to do retries for Read only I can look at reducing the "java-yness" of the builders.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
schmidt-sebastian: I looks like it is never changed though, at least I cannot find a place that modified this argument in this PR.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
schmidt-sebastian: 1. I am trying to get this SDK to take over some of the "standard retry" work that otherwise lives in every customer's application layer and ideally would offer a product that just works and irons out what I (personally) consider quirks of the backend.

2. This kind of saddens me but I think that is a valid point and means we should not offer retries.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +139,169 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+  @Deprecated
+  public static TransactionOptions create(@Nullable Executor executor, int numberOfAttempts) {
+    return readWriteOptionsBuilder()
+        .setExecutor(executor)
+        .setNumberOfAttempts(numberOfAttempts)
+        .build();
+  }
+
+  @Nonnull
+  public static ReadWriteOptionsBuilder readWriteOptionsBuilder() {
```
schmidt-sebastian, line 156: Just a random thought - feel free to ignore, but could this be `createReadWriteOptionsBuilder()`, which would kind of match the previous API (and break standard naming conventions for builders, which we could fix in the next major version!).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
schmidt-sebastian: Marking as resolved due to the discrepancy between retry/readTime.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
```diff
@@ -61,24 +63,34 @@
...
+        previousTransaction != null ? previousTransaction.transactionId : null;
   }
 
   /** Starts a transaction and obtains the transaction id. */
   ApiFuture<Void> begin() {
     BeginTransactionRequest.Builder beginTransaction = BeginTransactionRequest.newBuilder();
     beginTransaction.setDatabase(firestore.getDatabaseName());
 
-    if (previousTransactionId != null) {
+    if (originalTransactionOptions.getType() == TransactionOptionsType.READ_WRITE
```
schmidt-sebastian, line 85: Nit: https://engdoc.corp.google.com/eng/doc/devguide/java/practices/enums.md?cl=head#equals

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
BenWhitehead: Yup, I'll rename the builder factory methods.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
BenWhitehead: Will fix

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,161 @@
...
-/** Options specifying the behavior of Firestore Transactions. */
+/**
+ * Options specifying the behavior of Firestore Transactions.
+ *
+ * <p>A transaction in Firestore can be either read-write or read-only.
+ *
+ * <p>The default set of options is a read-write transaction with a max attempt count of 5. This
+ * attempt count can be customized via the {@link ReadWriteOptionsBuilder#setNumberOfAttempts(int)}
+ * method. A new instance of a builder can be created by calling {@link
+ * #createReadWriteOptionsBuilder()}.
```
schmidt-sebastian, line 39:
```suggestion
 * <p>The default set of options is a read-write transaction with a maximum number of 5 attempts. This
 * attempt count can be customized via the {@link ReadWriteOptionsBuilder#setNumberOfAttempts(int)}
 * method. A new instance of a builder can be created by calling {@link
 * #createReadWriteOptionsBuilder()}.
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,161 @@
...
+    this.readWrite = readWrite;
   }
 
+  /**
+   * @return the initial number of attempts a read-write transaction will be attempted
+   * @deprecated as of v2.0.0, only applicable to Read-Write transactions. Use {@link
+   *     ReadWriteOptions#getNumberOfAttempts()} instead
+   */
+  @Deprecated
+  @InternalApi
```
schmidt-sebastian, line 75: This should not be `InternalApi` if I am not mistaken.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +179,264 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+  @Deprecated
+  public static TransactionOptions create(@Nullable Executor executor, int numberOfAttempts) {
+    return createReadWriteOptionsBuilder()
+        .setExecutor(executor)
+        .setNumberOfAttempts(numberOfAttempts)
+        .build();
+  }
+
+  /**
+   * @return a new builder with default values applicable to configuring options for a read-write
```
schmidt-sebastian, line 196: s/builder/Builder ? (same below)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +179,264 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+
+    /** @return an instance of {@link TransactionOptions} from the values passed to this builder */
+    @Nonnull
+    public abstract TransactionOptions build();
+  }
+
+  /**
+   * A typesafe builder class representing those options that are applicable when configuring a
+   * transaction to be read-only. All methods function as "set" rather than returning a new copy
+   * with a value set on it.
```
schmidt-sebastian, line 250: I would drop "All methods function as "set" rather than returning a new copy with a value set on it." since this seems expected.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +179,264 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+    }
+
+    /** @return the currently set value that will be used as the readTime. */
+    @Nullable
+    public TimestampOrBuilder getReadTime() {
+      return readTime;
+    }
+
+    /**
+     * Specify to read documents at the given time. This may not be more than 60 in the past from
```
schmidt-sebastian, line 267: s/60/60 seconds/

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +179,264 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+      final ReadOnly.Builder builder = ReadOnly.getDefaultInstance().toBuilder();
+      if (readTime != null) {
+        builder.setReadTime(readTime);
+      }
+      return builder.build();
+    }
+
+    /**
+     * Specify to read documents at the given time. This may not be more than 60 in the past from
+     * when the request is processed by the server.
```
schmidt-sebastian, line 370: This should be rewritten to indicate that this is the getter.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +179,264 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+     *
+     * @return The specific time to read documents at. Must not be older than 60 seconds. A null
+     *     value means read most up to date data.
```
schmidt-sebastian, line 373:
```suggestion
     * @return The specific time to read documents at. A null
     *     value means read most up to date data.
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -83,10 +179,264 @@ public static TransactionOptions create(@Nonnull Executor executor) {
...
+    private ReadWriteOptions(int numberOfAttempts) {
+      this.numberOfAttempts = numberOfAttempts;
+    }
+
+    ReadWrite toProto() {
+      return PROTO;
+    }
+
+    /**
+     * Specify the max number of attempts a transaction will be attempted before resulting in an
```
schmidt-sebastian, line 416:
```suggestion
     * Returns the maximum number of times a transaction will be attempted before resulting in an
```

schmidt-sebastian:
Basically LGTM, pending our discussion on chat on whether we can simplify the TransactionOptions getters.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
```diff
@@ -78,11 +85,18 @@ Transaction wrapResult(ApiFuture<WriteResult> result) {
...
       beginTransaction
           .getOptionsBuilder()
           .getReadWriteBuilder()
           .setRetryTransaction(previousTransactionId);
+    } else if (TransactionOptionsType.READ_ONLY.equals(transactionOptions.getType())
+        && transactionOptions.getReadTime() != null) {
+      beginTransaction
+          .getOptionsBuilder()
+          .getReadOnlyBuilder()
+          .setReadTime(transactionOptions.getReadTime());
```
schmidt-sebastian, line 99: I think we must call `beginTransaction.getOptionsBuilder().getReadOnlyBuilder()` for read-only transaction even if we don't have a read time. You can probably just drop ` && transactionOptions.getReadTime() != null` (and maybe also `&& previousTransactionId != null`).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,133 @@
...
+   * @return The type of transaction this represents. Either read-only or read-write.
+   */
+  @Nonnull
+  public TransactionOptionsType getType() {
+    return type;
+  }
+
+  /**
+   * A {@link Timestamp} specifying the time documents are to be read at. If null, the server will
+   * read documents at the most up to date available. If nonnull, the specified {@code Timestamp}
```
schmidt-sebastian, line 93: s/nonnull/non-null ?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,133 @@
...
+   * may not be more than 60 seconds in the past (evaluated when the request is processed by the
+   * server).
+   *
+   * @return The specific time to read documents at. A null value means read most up to date data.
```
schmidt-sebastian, line 97:
```suggestion
   * @return The specific time to read documents at. A null value means reading the most up to date data.
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
```diff
@@ -16,65 +16,133 @@
...
+   */
+  @Nullable
+  public Timestamp getReadTime() {
+    if (TransactionOptionsType.READ_ONLY.equals(type)) {
+      return readTime;
+    } else {
+      return null;
+    }
```
schmidt-sebastian, line 105: This if statement is technically not needed, but it might help others understand how this works.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
BenWhitehead: I'll rename it before merging

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
BenWhitehead: Good catch. I did need to handle the else branch here that was previously taken care of by the options class. I've updated the code to handle both cases.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionOptions.java
BenWhitehead: Yeah, I kept it in here for clarity sake. I'll add a comment next to it stating as much.

PR #323, opened 2020-08-07
feat: add BulkWriter

thebrianchen:


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
```diff
@@ -0,0 +1,53 @@
...
+package com.google.cloud.firestore;
+
+import com.google.cloud.Timestamp;
+import javax.annotation.Nullable;
+
+/**
+ * A BatchWriteResult wraps the write time and status returned by Firestore when making
+ * BatchWriteRequests.
+ */
+public final class BatchWriteResult {
```
BenWhitehead, line 26: This class appears to be an internal class so we can scope down the visibility and annotate as such.

```suggestion
@IntenralApi
final class BatchWriteResult {
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -0,0 +1,763 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/** Used to represent a batch on the BatchQueue. */
+class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
```
BenWhitehead, line 45: It is very unusual to have multiple classes at the top level of a java file.

Can we move BulkCommitBatch to its own file in this same package?

BenWhitehead: Or, if it's only needed by the BatchWriter class, we can make it a static inner class of BatchWriter and keep it in this file, but moved.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -0,0 +1,763 @@
...
+    return result;
+  }
+
+  @Override
+  boolean allowDuplicateDocs() {
+    return false;
+  }
+}
+
+public class BulkWriter {
```
BenWhitehead, line 84: Let's make this class `final` so people don't extend it.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -0,0 +1,763 @@
...
+   *
+   * <p>Returns an ApiFuture that completes when there are no more pending writes. The ApiFuture
+   * will never error. Calling this method will send all requests. The ApiFuture completes
+   * immediately if there are no pending writes.
+   *
+   * @return An ApiFuture that completes when all enqueued writes up to this point have been
+   *     committed.
+   */
+  @Nonnull
+  public ApiFuture<Void> close() {
```
BenWhitehead, line 524: It appears this method is only used in tests right now. Can we scope it to package private, or is the intent that users will call this themselves?

There are a couple of interfaces in java, notably the general purpose [`java.lang.AutoClosable`](https://docs.oracle.com/javase/8/docs/api/index.html?java/lang/AutoCloseable.html#close--) and the IO specific [`java.io.Closable`](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html#close--) which already have defined method signatures for `close`, specifically `void close() ...` if we ship this public signature then we would prevent the ability to use a BulkWriter in a [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) context.

If we do want to keep the method public, can we rename it to something like `completeAll` or `closeAll` to remove the name collision?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
BenWhitehead, line 84: Following the comment about `close()`, if we do expect users to call close, can we also make this class implement AutoClosable, and define the close method as waiting for all futures to complete before yielding and swallowing the results.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOptions.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore;
+
+import javax.annotation.Nonnull;
+
+/** Options used to disable request throttling in BulkWriter. */
+public class BulkWriterOptions {
```
BenWhitehead, line 22: Can we make this class final to prevent subclassing?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
```diff
@@ -38,6 +38,10 @@
  */
 public final class Transaction extends UpdateBuilder<Transaction> {
 
+  Transaction wrapResult(ApiFuture<WriteResult> result) {
+    return this;
+  }
+
```
BenWhitehead, line 44: nit: package-public method are generally found after public method in a class, not before the static fields.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -43,17 +52,62 @@
  * Abstract class that collects and bundles all write operations for {@link Transaction} and {@link
  * WriteBatch}.
  */
-public abstract class UpdateBuilder<T extends UpdateBuilder> {
+public abstract class UpdateBuilder<T> {
```
BenWhitehead, line 55: Please add the `@InternalExtensionOnly` (from `com.google.api.core`) annotation to this class so we're covered should we need to modify anything

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
thebrianchen: done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: Moved to a separate file.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: Done.

thebrianchen: Implemented AutoClosable in BulkWriter and changed to synchronous method.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOptions.java
thebrianchen: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Transaction.java
thebrianchen: Moved further down.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
thebrianchen: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: Follow up from conversation:
"AutoClosable already dictates the method signature of close as `public void close() throws Exception` so, the existing `ApiFuture<Void> close()` will collide with it and result in a compilation error."

Solution: Change `close()`to synchronous method that blocks on the writes, which will allow the override to work and still support AutoClosable.

PR #337, opened 2020-08-17
fix: add support to deserialize to custom Lists and Maps

### LINKED ISSUE #318: Can't deserialize a class extending ArrayList
#### Environment details

1. Specify the API at the beginning of the title. For example, "BigQuery: ...").
   com.google.cloud:google-cloud-firestore:1.33.0
2. OS type and version:
  Windows 10
3. Java version:
  Java 11
4. firestore version(s):
1.33.0
#### Steps to reproduce

  1. Load a class having a property extending the ArrayList class (InternalFieldDefinition in my example)

#### Code example

```java
package io.test;

import java.util.ArrayList;

public class InternalFieldPath extends ArrayList<InternalMappingField> {

    public InternalFieldPath() {
        super();
    }

}
```
```java

package io.test;

public class InternalFieldDefinition extends InternalModel {

    private String name;
    private InternalFieldPath path;

}
```

#### Stack trace
```
java.lang.RuntimeException: Could not deserialize object. Can't convert object of type java.util.ArrayList to type io.blendeez.common.domain.model.internal.InternalFieldPath (found in field 'fieldDefinitions.[0].sourceFieldPath')
	at com.google.cloud.firestore.CustomClassMapper.deserializeError(CustomClassMapper.java:563)
        at com.google.cloud.firestore.CustomClassMapper.convertBean(CustomClassMapper.java:544)
```

#### Any additional information below
It seems the case where a bean extends an ArrayList is not properly handled in CustomClassMapper (only Map):

```java
  private static <T> T convertBean(Object o, Class<T> clazz, DeserializeContext context) {
    BeanMapper<T> mapper = loadOrCreateBeanMapperForClass(clazz);
    if (o instanceof Map) {
      return mapper.deserialize(expectMap(o, context), context);
    } else {
      throw deserializeError(
          context.errorPath,
          "Can't convert object of type " + o.getClass().getName() + " to type " + clazz.getName());
    }
  }
```

FYI while switching the class attribute to a list it works:

```java
  public class InternalFieldDefinition {

    private String name;
    private List<InternalMappingField> path;

}
```

suraj-qlogic:
Fixes #318

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
@@ -266,7 +266,18 @@ class CustomClassMapper {
       Type genericType = type.getActualTypeArguments()[0];
       if (o instanceof List) {
         List<Object> list = (List<Object>) o;
-        List<Object> result = new ArrayList<>(list.size());
+        List<Object> result = null;
+        try {
+          result =
+              (rawType == List.class)
+                  ? new ArrayList<>(list.size())
+                  : (List<Object>) rawType.getDeclaredConstructor().newInstance();
+        } catch (InstantiationException
+            | IllegalAccessException
+            | NoSuchMethodException
+            | InvocationTargetException e) {
+          throw new RuntimeException(e);
+        }
         for (int i = 0; i < list.size(); i++) {
           result.add(
               deserializeToType(
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
@@ -279,8 +279,7 @@ class CustomClassMapper {
           throw deserializeError(
               context.errorPath,
               String.format(
-                  "Unable to deserialize to the %s type: %s",
-                  rawType.getSimpleName(), e.toString()));
+                  "Unable to deserialize to %s: %s", rawType.getSimpleName(), e.toString()));
         }
         for (int i = 0; i < list.size(); i++) {
           result.add(
@@ -315,7 +314,7 @@ class CustomClassMapper {
         throw deserializeError(
             context.errorPath,
             String.format(
-                "Unable to deserialize to the %s type: %s", rawType.getSimpleName(), e.toString()));
+                "Unable to deserialize to %s: %s", rawType.getSimpleName(), e.toString()));
       }
       for (Map.Entry<String, Object> entry : map.entrySet()) {
         result.put(
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
@@ -266,7 +266,7 @@ class CustomClassMapper {
       Type genericType = type.getActualTypeArguments()[0];
       if (o instanceof List) {
         List<Object> list = (List<Object>) o;
-        List<Object> result = null;
+        List<Object> result;
         try {
           result =
               (rawType == List.class)
@@ -276,7 +276,11 @@ class CustomClassMapper {
             | IllegalAccessException
             | NoSuchMethodException
             | InvocationTargetException e) {
-          throw new RuntimeException(e);
+          throw deserializeError(
+              context.errorPath,
+              String.format(
+                  "Unable to deserialize to the %s type: %s",
+                  rawType.getSimpleName(), e.toString()));
         }
         for (int i = 0; i < list.size(); i++) {
           result.add(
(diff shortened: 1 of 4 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
```diff
@@ -266,7 +266,18 @@ static Object convertToPlainJavaTypes(Object object) {
...
+        try {
+          result =
+              (rawType == List.class)
+                  ? new ArrayList<>(list.size())
+                  : (List<Object>) rawType.getDeclaredConstructor().newInstance();
+        } catch (InstantiationException
+            | IllegalAccessException
+            | NoSuchMethodException
+            | InvocationTargetException e) {
+          throw new RuntimeException(e);
```
schmidt-sebastian, line 279: Just FYI: While uncommon, I think this method would now throw if called on an type that represents an subinterface of List, e.g.:

interace MyList extends List {
}

class Pojo {
  MyList data;
}
```

I am however not 100% certain that is the case, but based on my reading of https://stackoverflow.com/questions/36564041/getdeclaredconstructor-on-an-interface, this would throw an NoSuchMethodException, whereas before it would just create an ArrayList (albeit of the wrong type). I don't know if there is a solution that doesn't involve creating a list of the wrong type, and if there isn't, we can probably leave this logic as is.

**Comment by suraj-qlogic** [comment]
@schmidt-sebastian ,Yes you are right.Whenever I try to create an instance of interface it's produce NoSuchMethodException. Using reflection api won't provide support for creating an instance of interface.So alternative i think to produce deserialization error.

```
try {
	  result =(rawType == List.class)
		  ? new ArrayList<>(list.size())
		  : (List<Object>) rawType.getDeclaredConstructor().newInstance();
     }catch (InstantiationException
	    | IllegalAccessException
	    | NoSuchMethodException
	    | InvocationTargetException e) {
	  throw deserializeError(context.errorPath,String.format("Deserializing values to %s interface 
          is not supported : %s", rawType.getSimpleName(),e.toString()));
    }
```

**Comment by schmidt-sebastian** [comment]
> @schmidt-sebastian ,Yes you are right.Whenever I try to create an instance of interface it's produce NoSuchMethodException. Using reflection api won't provide support for creating an instance of interface.So alternative i think to produce deserialization error.

That error message LGTM. Do you want to update the PR and also add Map support?

**Comment by suraj-qlogic** [comment]
@schmidt-sebastian ,Thanks i will update this ASAP.

**Comment by dmitry-fa** [comment]
a nit: 
> "Deserializing values to %s interface is not supported : %s"

rawType is not necessary to be an interface to cause a deserialization problem, it could be an abstract class, a class without the default constructor, or with the private default constructor.

I would rephrase: `"Unable to deserialize to the %s type: %s"`

File: google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java [hunk 2]
```diff
@@ -266,7 +266,22 @@ static Object convertToPlainJavaTypes(Object object) {
...
+                  ? new ArrayList<>(list.size())
+                  : (List<Object>) rawType.getDeclaredConstructor().newInstance();
+        } catch (InstantiationException
+            | IllegalAccessException
+            | NoSuchMethodException
+            | InvocationTargetException e) {
+          throw deserializeError(
+              context.errorPath,
+              String.format(
+                  "Unable to deserialize to the %s type: %s",
```
schmidt-sebastian, line 282:
```suggestion
                  "Unable to deserialize to %s: %s",
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
```diff
@@ -287,7 +302,21 @@ static Object convertToPlainJavaTypes(Object object) {
...
+                ? new HashMap<String, Object>()
+                : (HashMap<String, Object>) rawType.getDeclaredConstructor().newInstance();
+      } catch (InstantiationException
+          | IllegalAccessException
+          | NoSuchMethodException
+          | InvocationTargetException e) {
+        throw deserializeError(
+            context.errorPath,
+            String.format(
+                "Unable to deserialize to the %s type: %s", rawType.getSimpleName(), e.toString()));
```
schmidt-sebastian, line 318:
```suggestion
                  "Unable to deserialize to %s: %s",
```

schmidt-sebastian (approved):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
suraj-qlogic: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
suraj-qlogic: Done

PR #350, opened 2020-08-31
feat: add support for != and NOT_IN queries

thebrianchen:
Porting from [node](https://github.com/googleapis/nodejs-firestore/pull/1292)

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
@@ -25,6 +25,8 @@ import static com.google.firestore.v1.StructuredQuery.FieldFilter.Operator.GREAT
 import static com.google.firestore.v1.StructuredQuery.FieldFilter.Operator.IN;
 import static com.google.firestore.v1.StructuredQuery.FieldFilter.Operator.LESS_THAN;
 import static com.google.firestore.v1.StructuredQuery.FieldFilter.Operator.LESS_THAN_OR_EQUAL;
+import static com.google.firestore.v1.StructuredQuery.FieldFilter.Operator.NOT_EQUAL;
+import static com.google.firestore.v1.StructuredQuery.FieldFilter.Operator.NOT_IN;
 
 import com.google.api.core.ApiFuture;
 import com.google.api.core.InternalExtensionOnly;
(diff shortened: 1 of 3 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -492,6 +494,48 @@ public Query whereEqualTo(@Nonnull FieldPath fieldPath, @Nullable Object value)
     }
   }
 
+  /**
+   * Creates and returns a new Query with the additional filter that documents must contain the
+   * specified field and the value does not equal the specified value.
```
schmidt-sebastian, line 499: This might be better (but no need to change unless we update all of the documentation):

```suggestion
   * specified field and its value does not equal the specified value.
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -732,6 +778,46 @@ public Query whereIn(@Nonnull FieldPath fieldPath, @Nonnull List<? extends Objec
     return whereHelper(fieldPath, IN, values);
   }
 
+  /**
+   * Creates and returns a new Query with the additional filter that documents must contain the
+   * specified field and the value does not equal any of the values from the provided list.
+   *
+   * <p>A Query can have only one whereNotIn() filter, and it cannot be combined with
```
schmidt-sebastian, line 785:
```suggestion
   * <p>A Query can have only one whereNotIn() filter and it cannot be combined with
```

(I think)

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1233,6 +1266,43 @@ public void inQueriesWithDocumentId() throws Exception {
...
+    assertEquals(asList("b", "a", "c", "d", "e", "f"), querySnapshotToIds(querySnapshot));
+
+    List<Object> nullArray = new ArrayList<>();
+    nullArray.add(null);
+    querySnapshot = randomColl.whereNotIn("zip", nullArray).get().get();
+    assertEquals(new ArrayList<>(), querySnapshotToIds(querySnapshot));
+  }
+
+  @Test
+  public void inQueriesWithDocumentId() throws Exception {
```
schmidt-sebastian, line 1292:
```suggestion
  public void notInQueriesWithDocumentId() throws Exception {
```

schmidt-sebastian (approved):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
schmidt-sebastian: Note that this and the next comment apply to multiple places in this file.

PR #377, opened 2020-09-22
chore: exclude Bundle Proto from Synthtool run

schmidt-sebastian:
Fixes https://github.com/googleapis/java-firestore/issues/371


schmidt-sebastian:
cc @wu-hui

BenWhitehead (approved):

PR #394, opened 2020-10-01
fix: bulkWriter: writing to the same doc doesn't create a new batch

thebrianchen:
Porting from [node](https://github.com/googleapis/nodejs-firestore/pull/1298).

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
@@ -26,21 +26,18 @@ import javax.annotation.Nullable;
  */
 @InternalApi
 public final class BatchWriteResult {
-  private final DocumentReference documentReference;
+  private final String key;
   @Nullable private final Timestamp writeTime;
   @Nullable private final Exception exception;
 
-  BatchWriteResult(
-      DocumentReference documentReference,
-      @Nullable Timestamp timestamp,
-      @Nullable Exception exception) {
-    this.documentReference = documentReference;
+  BatchWriteResult(String key, @Nullable Timestamp timestamp, @Nullable Exception exception) {
+    this.key = key;
     this.writeTime = timestamp;
     this.exception = exception;
   }
 
-  public DocumentReference getDocumentReference() {
-    return documentReference;
+  public String getKey() {
+    return key;
   }
 
   @Nullable
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
@@ -26,19 +26,12 @@ import javax.annotation.Nullable;
  */
 @InternalApi
 final class BatchWriteResult {
-  private final String documentKey;
   @Nullable private final Timestamp writeTime;
   @Nullable private final Exception exception;
 
-  BatchWriteResult(
-      String documentKey, @Nullable Timestamp timestamp, @Nullable Exception exception) {
-    this.documentKey = documentKey;
+  BatchWriteResult(@Nullable Timestamp timestamp, @Nullable Exception exception) {
     this.writeTime = timestamp;
     this.exception = exception;
-  }
-
-  public String getDocumentKey() {
-    return documentKey;
   }
 
   @Nullable
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
@@ -25,19 +25,20 @@ import javax.annotation.Nullable;
  * BatchWriteRequests.
  */
 @InternalApi
-public final class BatchWriteResult {
-  private final String key;
+final class BatchWriteResult {
+  private final String documentKey;
   @Nullable private final Timestamp writeTime;
   @Nullable private final Exception exception;
 
-  BatchWriteResult(String key, @Nullable Timestamp timestamp, @Nullable Exception exception) {
-    this.key = key;
+  BatchWriteResult(
+      String documentKey, @Nullable Timestamp timestamp, @Nullable Exception exception) {
+    this.documentKey = documentKey;
     this.writeTime = timestamp;
     this.exception = exception;
   }
 
-  public String getKey() {
-    return key;
+  public String getDocumentKey() {
+    return documentKey;
   }
 
   @Nullable
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
@@ -18,9 +18,6 @@ package com.google.cloud.firestore;
 
 import com.google.api.core.ApiFuture;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Predicate;
-import com.google.common.collect.FluentIterable;
-import java.util.Set;
 
 /** Used to represent a batch on the BatchQueue. */
 class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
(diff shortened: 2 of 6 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
```diff
@@ -26,21 +26,18 @@
  */
 @InternalApi
 public final class BatchWriteResult {
-  private final DocumentReference documentReference;
+  private final String key;
```
schmidt-sebastian, line 29: It looks like `BatchWriteResult` could be package-private, which would avoid the CLIRR violation.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
```diff
@@ -29,21 +26,13 @@
...
-            .filter(
-                new Predicate<WriteOperation>() {
-                  @Override
-                  public boolean apply(WriteOperation writeOperation) {
-                    return docsToRetry.contains(writeOperation.documentReference);
-                  }
-                })
-            .toList());
+
+    // Create a new BulkCommitBatch containing only the indexes from the provided indexes to retry.
```
schmidt-sebastian, line 32: Hm. This is different from Node, but likely better. Should we update the Node SDK to match? This might make porting simpler.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
schmidt-sebastian, line 29: Also s/key/documentKey since this is not apparent from the classes context (or documentPath - see below).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
```diff
@@ -26,21 +26,18 @@
...
-    this.documentReference = documentReference;
+  BatchWriteResult(String key, @Nullable Timestamp timestamp, @Nullable Exception exception) {
+    this.key = key;
     this.writeTime = timestamp;
     this.exception = exception;
   }
 
-  public DocumentReference getDocumentReference() {
-    return documentReference;
+  public String getKey() {
```
schmidt-sebastian, line 39: s/getKey/getDocumentKey

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -65,14 +66,30 @@
...
   }
 
+  /**
+   * Used to represent a pending write operation.
+   *
+   * <p>Contains a pending write's index, document path, and the corresponding result.
+   */
+  static class PendingOperation {
+    int writeBatchIndex;
+    String key;
```
schmidt-sebastian, line 76: s/key/documentKey (or documentPath)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -736,29 +749,51 @@ BatchState getState() {
...
 
-  boolean hasDocument(DocumentReference documentReference) {
-    return pendingOperations.containsKey(documentReference);
+  List<String> getPendingDocumentPaths() {
+    return FluentIterable.from(pendingOperations)
+        .transform(
+            new Function<PendingOperation, String>() {
+              @Override
+              public String apply(PendingOperation input) {
+                return input.key;
```
schmidt-sebastian, line 758: You should settle on either "key" or "path". I can see an argument for key (in that our DocumentReference.getPath method is relative to the root), but we usually use path to represent the full document path in its Proto representation. I think "key" is exactly that here.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -736,29 +749,51 @@ BatchState getState() {
...
+  List<SettableApiFuture<WriteResult>> getPendingFutures() {
+    return FluentIterable.from(pendingOperations)
+        .transform(
+            new Function<PendingOperation, SettableApiFuture<WriteResult>>() {
+              @Override
+              public SettableApiFuture<WriteResult> apply(PendingOperation input) {
+                return input.future;
+              }
+            })
+        .toList();
```
schmidt-sebastian, line 785: This code almost looks like the future!

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -736,29 +749,51 @@ BatchState getState() {
...
 
   private ApiFuture<WriteResult> processOperation(DocumentReference documentReference) {
-    Preconditions.checkState(
-        allowDuplicateDocs() || !pendingOperations.containsKey(documentReference),
-        "Batch should not contain writes to the same document");
     Preconditions.checkState(state == BatchState.OPEN, "Batch should be OPEN when adding writes");
     SettableApiFuture<WriteResult> result = SettableApiFuture.create();
-    pendingOperations.put(documentReference, result);
+    pendingOperations.add(
+        new PendingOperation(getPendingOperationCount(), documentReference.getPath(), result));
```
schmidt-sebastian, line 796:
```suggestion
    int nextBatchIndex = getPendingOperationCount();
    pendingOperations.add(
        new PendingOperation(nextBatchIndex, documentReference.getPath(), result));
```

Optional, but it does remove the need to understand the execution order.

schmidt-sebastian (commented):
Basically LGTM. Some naming nits.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
thebrianchen: good point, done.

thebrianchen: Thanks to IntelliJ, realized this field is no longer needed. Will remove in Node as well.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
thebrianchen: done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
thebrianchen: Funny how having more tools at your disposal does not translate to writing better code. Will create a small node PR for that later.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
thebrianchen: done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
thebrianchen: I prefer `documentKey`, for the reasons you mentioned, and b/c it matches better with `key` from Node.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
thebrianchen: done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BatchWriteResult.java
thebrianchen: It added another CLIRR violation for changing to package private, but I changed the rule.

schmidt-sebastian (approved):

BenWhitehead (approved):

PR #406, opened 2020-10-07
chore: Remove update.sh and generate bundle.proto code from googleapis

wu-hui:


schmidt-sebastian (approved):

schmidt-sebastian:
One nit: This is probably a "chore" rather than a "fix" since it does not need to be released to customers.

wu-hui:
> One nit: This is probably a "chore" rather than a "fix" since it does not need to be released to customers.

Done.

PR #418, opened 2020-10-14
fix: verify page count before invoking QueryPartition RPC

schmidt-sebastian:
Adding a manual verification for the partition count since we are requesting one less partition from the backend, which creates a confusing error message if the user uses a negative (or zero) partition count.


schmidt-sebastian:
@BenWhitehead Do you happen to know if there is any sample code I could look at that shows how to mock a PartitionQueryPagedResponse?

BenWhitehead (approved):

PR #426, opened 2020-10-19
fix: retry Query streams

schmidt-sebastian:
This restarts query streams that fail with a retryable error based on the last document received.

Port of https://github.com/googleapis/nodejs-firestore/pull/1116

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
@@ -908,7 +908,7 @@ public class QueryTest {
   }
 
   @Test
-  public void retriesAfterError() throws Exception {
+  public void retriesAfterRetryableError() throws Exception {
     final boolean[] returnError = new boolean[] {true};
 
     doAnswer(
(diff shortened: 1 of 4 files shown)


schmidt-sebastian:
cc @crwilcox

google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
```diff
@@ -902,6 +907,72 @@ public void onCompleted() {
...
+    final boolean[] returnError = new boolean[] {true};
+
+    doAnswer(
+            new Answer<RunQueryResponse>() {
+              public RunQueryResponse answer(InvocationOnMock invocation) throws Throwable {
+                if (returnError[0]) {
+                  returnError[0] = false;
+                  return queryResponse(
+                          FirestoreException.serverRejected(
+                              Status.DEADLINE_EXCEEDED, "Simulated test failure"),
```
wu-hui, line 921: I wonder if it makes sense to add a test to verify the no-retry behavior where it is not suppose to re-try.

wu-hui (approved):

PR #458, opened 2020-11-09
feat: add error and success handlers to BulkWriter

thebrianchen:
Porting of several node PRs: [1](https://github.com/googleapis/nodejs-firestore/pull/1315), [2](https://github.com/googleapis/nodejs-firestore/pull/1352), [3](https://github.com/googleapis/nodejs-firestore/pull/1354).

Another fix I made was reducing the `MAX_BATCH_SIZE` from 500 to 20. I had to do this after refactoring UpdateBuilder, since the previous code imposed the same limits on WriteBatch/Transactions, causing the system tests to fail.

I've tried to order the commits to make it more reviewable, but after finishing it up the first time, I went back and did more cleanup.


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
```diff
@@ -16,32 +16,164 @@
...
+  ApiFuture<WriteResult> wrapResult(DocumentReference documentReference) {
+    return processLastOperation(documentReference);
+  }
+
+  /**
+   * Commits all pending operations to the database and verifies all preconditions.
+   *
+   * <p>The writes in the batch are not applied atomically and can be applied out of order.
+   */
+  ApiFuture<List<BatchWriteResult>> bulkCommit() {
```
thebrianchen, line 66: I tried extracting a `commitHelper()` from `bulkCommit()` and `commit()`, but couldn't figure out how to properly parameterize everything.

Aside from reducing code duplication, the main reason I want to do this is because moving this over requires a getter for `writes` and exposing `committed` as `protected`. If we could move the logic for these parts into `UpdateBuilder`, we could make both of those variables private.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -515,70 +756,117 @@ private void verifyNotClosed() {
...
+   *   bulkWriter.addWriteResultListener(
+   *         new WriteResultCallback() {
+   *           public void onResult(DocumentReference documentReference, WriteResult result) {
+   *             System.out.println(
+   *                 "Successfully executed write on document: "
+   *                     + documentReference
+   *                     + " at: "
+   *                     + result.getUpdateTime());
+   *           }
+   *         });
```
schmidt-sebastian, line 773: You should probably just use Lamdba syntax here.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -515,70 +756,117 @@ private void verifyNotClosed() {
...
+   *   bulkWriter.addWriteErrorListener(
+   *         new WriteErrorCallback() {
+   *           if (error.getStatus() == Status.UNAVAILABLE
+   *                 && error.getFailedAttempts() < MAX_RETRY_ATTEMPTS) {
+   *               return true;
+   *             } else {
+   *               System.out.println("Failed write at document: " + error.getDocumentReference());
+   *               return false;
+   *             }
+   *         });
```
schmidt-sebastian, line 801: You should probably use Lambda syntax here. If you stick with the current example, you need to add the function signature.

schmidt-sebastian (commented):
This looks really good!

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterError.java
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore;
+
+import com.google.cloud.firestore.BulkWriter.OperationType;
+import io.grpc.Status;
+
+/** The error thrown when a BulkWriter operation fails. */
+class BulkWriterError extends Exception {
```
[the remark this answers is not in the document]
thebrianchen: The current BulkWriter hasn't been released publicly yet, so I haven't made this a public class yet.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -352,6 +356,325 @@ public void sendWritesToDifferentDocsInSameBatch() throws Exception {
...
+                failedResponse(Code.INTERNAL_VALUE));
+          }
+        };
+    responseStubber.initializeStub(batchWriteCapture, firestoreMock);
+
+    final boolean[] errorListenerCalled = {false};
+    bulkWriter.addWriteErrorListener(
+        new WriteErrorCallback() {
+          public boolean shouldRetryListener(BulkWriterError error) {
+            errorListenerCalled[0] = true;
```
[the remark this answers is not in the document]
thebrianchen: done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -352,6 +356,325 @@ public void sendWritesToDifferentDocsInSameBatch() throws Exception {
...
+                batchWrite(set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc1")),
+                failedResponse(Code.INTERNAL_VALUE));
+          }
+        };
+    responseStubber.initializeStub(batchWriteCapture, firestoreMock);
+
+    bulkWriter.addWriteErrorListener(
+        new WriteErrorCallback() {
+          public boolean shouldRetryListener(BulkWriterError error) {
+            throw new NullPointerException("Test code threw NullPointerException");
```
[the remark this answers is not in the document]
thebrianchen: done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -352,6 +356,325 @@ public void sendWritesToDifferentDocsInSameBatch() throws Exception {
...
+                batchWrite(set(LocalFirestoreHelper.SINGLE_FIELD_PROTO, "coll/doc1")),
+                successResponse(1));
+          }
+        };
+    responseStubber.initializeStub(batchWriteCapture, firestoreMock);
+
+    bulkWriter.addWriteResultListener(
+        new WriteResultCallback() {
+          public void onResult(DocumentReference documentReference, WriteResult result) {
+            throw new NullPointerException("Test code threw NullPointerException");
```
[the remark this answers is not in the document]
thebrianchen: done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -442,22 +575,99 @@
...
+          public ApiFuture<WriteResult> apply(FirestoreException exception) throws BulkWriterError {
+            BulkWriterError bulkWriterError =
+                new BulkWriterError(
+                    exception.getStatus(),
+                    exception.getMessage(),
+                    documentReference,
+                    operationType,
+                    failedAttempts);
+
+            boolean shouldRetry = errorListener.shouldRetryListener(bulkWriterError);
```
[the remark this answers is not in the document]
thebrianchen: Done here and in the success callback.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -609,97 +896,45 @@ public void run() {
...
-                  // If the BatchWrite RPC fails, map the exception to each individual result.
-                  for (int i = 0; i < batch.getPendingDocumentPaths().size(); ++i) {
-                    results.add(new BatchWriteResult(null, exception));
-                  }
-                  return ApiFutures.immediateFuture(results);
+  private ApiFuture<Void> bulkCommit(final BulkCommitBatch batch) {
+    return ApiFutures.transformAsync(
+        ApiFutures.catchingAsync(
+            batch.bulkCommit(),
+            Exception.class,
```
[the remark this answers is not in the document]
thebrianchen: Based on Java [docs](https://docs.oracle.com/javase/7/docs/api/java/lang/RuntimeException.html#:~:text=RuntimeException%20is%20the%20superclass%20of,its%20subclasses%20are%20unchecked%20exceptions.), `RunTimeException` extends `Exception`. The main issue with using Throwable is that `BatchWriteResult` takes in `Exception`, and I would have to plumb that through.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -59,13 +61,15 @@
      * @param error The error object from the failed BulkWriter operation attempt.
      * @return Whether to retry the operation or not.
      */
-    boolean shouldRetryListener(BulkWriterError error);
+    boolean shouldRetryError(BulkWriterError error);
```
schmidt-sebastian, line 64: I am thinking more and more that this should be `onError`. Otherwise, users are going to ask us how they can get the error result and why we don't expose it in  `WriteResultCallback`. Note also that the term `result` is a bit ambiguous as it doesn't define whether "error" is a "result".

We should maybe also ask for a third opinion of whether the classes are "callbacks".  On Android, we seem to call the outermost class "Listener". The method inside is prefixed with "on" (see EventListener).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -608,8 +613,9 @@ public boolean shouldRetryListener(BulkWriterError error) {
         ApiFutures.transformAsync(
             executeWriteHelper(documentReference, operationType, operationCallback, 0),
             new ApiAsyncFunction<WriteResult, WriteResult>() {
-              public ApiFuture<WriteResult> apply(WriteResult result) {
-                successListener.onResult(documentReference, result);
+              public ApiFuture<WriteResult> apply(WriteResult result)
+                  throws ExecutionException, InterruptedException {
+                invokeUserSuccessCallback(documentReference, result).get();
```
schmidt-sebastian, line 618: This looks like it may block the GRPC thread while the user code is running. Instead, you should do one of the following:

- Run your code
- Run user code in user thread, followed by a continuation that dispatches back on your old exectur
- Finish running your own code

Realistically, though, no one is going to get mad if you run a little bit of your own code on the use executor (as long as it is non-blocking and doesn't throw). For that reason, you can likely just use the user executor when the callback is first invoked (e.g. replace the directExecutor call in `executeWrite()` if deemed safe).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -762,8 +806,7 @@ private void verifyNotClosed() {
    * <p>For example, see the sample code: <code>
    *   BulkWriter bulkWriter = firestore.bulkWriter();
    *   bulkWriter.addWriteResultListener(
-   *         new WriteResultCallback() {
-   *           public void onResult(DocumentReference documentReference, WriteResult result) {
+   *         (DocumentReference documentReference, WriteResult result -> {
```
schmidt-sebastian, line 809: Missing ")".

We should make sure these compile as users are going to copy them.

thebrianchen: Added. Checked that code compiles.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: Using userExecutor for `executeWrite()`, and got rid of the invokeCallback methods.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: Changed to `onError`. As for WriteResult, we discussed naming for the node implementation and decided on `WriteResult` over `WriteSuccess` since a `WriteResult` is returned from the operation.

Would it be too sadistic to ask Ben to take a look at this PR for Java code style/convention aftewards?

schmidt-sebastian: We should definitely ask Ben :)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: Summary of conversation w/ Ben: We should not block grpc or gax threads. Instead, we should be creating a thread on each firestore instance that awaits user callbacks.

Solution: Created a separate thread for waiting on the callback. Fixed async issue by fixing bug in `performFlush()`, where we were incorrectly resolving the `flushComplete` future early if no retries were scheduled.

thebrianchen:
Continued in #483.

PR #466, opened 2020-11-16
feat: Release bundles

wu-hui:
PR to release bundles. Will merge when we decide to release.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
@@ -169,6 +169,16 @@ public interface Firestore extends Service<FirestoreOptions>, AutoCloseable {
   WriteBatch batch();
 
   /**
+   * Returns a FirestoreBundle.Builder {@link FirestoreBundle.Builder} instance for the given bundle
+   * Id.
+   *
+   * @param bundleId If null, an random string will be used for the Id. The Id is used by the client
+   *     SDKs loading bundles to identify if a bundle has been loaded before.
+   */
+  @Nonnull
+  FirestoreBundle.Builder bundleBuilder(@Nullable String bundleId);
+
+  /**
    * Closes the gRPC channels associated with this instance and frees up their resources. This
    * method blocks until all channels are closed. Once this method is called, this Firestore client
    * is no longer usable.
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
@@ -170,8 +170,8 @@ public interface Firestore extends Service<FirestoreOptions>, AutoCloseable {
 
   /**
    * Returns a FirestoreBundle.Builder {@link FirestoreBundle.Builder} instance using an
-   * automatically generated ID. The ID is used by the client SDKs loading bundles to identify if a
-   * bundle has been loaded before.
+   * automatically generated bundle ID. When loaded on clients, client SDKs use the bundle ID and
+   * the timestamp associated with the built bundle to tell if it has been loaded already.
    */
   @Nonnull
   FirestoreBundle.Builder bundleBuilder();
@@ -180,8 +180,8 @@ public interface Firestore extends Service<FirestoreOptions>, AutoCloseable {
    * Returns a FirestoreBundle.Builder {@link FirestoreBundle.Builder} instance for the given bundle
    * ID.
    *
-   * @param bundleId The ID is used by the client SDKs loading bundles to identify if a bundle has
-   *     been loaded before. If null, a random string will be used for the ID.
+   * @param bundleId The ID of the bundle. When loaded on clients, client SDKs use this id and the
+   *     timestamp associated with the built bundle to tell if it has been loaded already.
    */
   @Nonnull
   FirestoreBundle.Builder bundleBuilder(@Nonnull String bundleId);
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
@@ -184,7 +184,7 @@ public interface Firestore extends Service<FirestoreOptions>, AutoCloseable {
    *     been loaded before. If null, a random string will be used for the ID.
    */
   @Nonnull
-  FirestoreBundle.Builder bundleBuilder(@Nullable String bundleId);
+  FirestoreBundle.Builder bundleBuilder(@Nonnull String bundleId);
 
   /**
    * Closes the gRPC channels associated with this instance and frees up their resources. This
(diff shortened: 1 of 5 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -168,6 +168,16 @@ void getAll(
   @Nonnull
   WriteBatch batch();
 
+  /**
+   * Returns a FirestoreBundle.Builder {@link FirestoreBundle.Builder} instance for the given bundle
+   * Id.
```
schmidt-sebastian, line 173: s/Id/ID

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -168,6 +168,16 @@ void getAll(
...
+   *
+   * @param bundleId If null, an random string will be used for the Id. The Id is used by the client
+   *     SDKs loading bundles to identify if a bundle has been loaded before.
```
schmidt-sebastian, line 176: s/Id/ID and swap both sentences.

If a null bundle ID is how me mainly expect the bundles to be used, should we add an overload that doesn't require an explicit `null`?

BenWhitehead:
New methods on the Firestore interface will need ignore rules added to [`google-cloud-firestore/clirr-ignored-differences.xml`](https://github.com/googleapis/java-firestore/blob/master/google-cloud-firestore/clirr-ignored-differences.xml). An example of how to allow a new method on the interface can be see in [this rule](https://github.com/googleapis/java-firestore/blob/master/google-cloud-firestore/clirr-ignored-differences.xml#L27-L31).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
wu-hui: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
wu-hui: Yep, that is better.

I also realized we provided no way for the client to get the bundle id if it is auto-id, they probably want to use it to manage their bundle serving. I added a `getId` to builder. Will pass a note to the council again.

wu-hui (commented):
Also add to clirr ignore file.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -168,6 +168,24 @@ void getAll(
...
+   */
+  @Nonnull
+  FirestoreBundle.Builder bundleBuilder();
+
+  /**
+   * Returns a FirestoreBundle.Builder {@link FirestoreBundle.Builder} instance for the given bundle
+   * ID.
+   *
+   * @param bundleId The ID is used by the client SDKs loading bundles to identify if a bundle has
+   *     been loaded before. If null, an random string will be used for the ID.
```
schmidt-sebastian, line 184: s/an/a

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
schmidt-sebastian: Thought on adding `bundleBuilder()` with no arguments? I believe this would be the main way to interact with bundles, and we should make it as easy as possible.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
wu-hui: Done.

schmidt-sebastian:
@wu-hui What are your thoughts on the overload that doesn't take a bundle ID?

wu-hui:
> @wu-hui What are your thoughts on the overload that doesn't take a bundle ID?

Sorry I meant to reply your earlier comment.

I think it is added (https://github.com/googleapis/java-firestore/pull/466/files#diff-b9277c291d91b6930b9a93135a3090afa8aebdacc3c728dec58f858db2bc7b95R177), were you looking at an older commit?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -168,6 +168,24 @@ void getAll(
...
+
+  /**
+   * Returns a FirestoreBundle.Builder {@link FirestoreBundle.Builder} instance for the given bundle
+   * ID.
+   *
+   * @param bundleId The ID is used by the client SDKs loading bundles to identify if a bundle has
+   *     been loaded before. If null, a random string will be used for the ID.
+   */
+  @Nonnull
+  FirestoreBundle.Builder bundleBuilder(@Nullable String bundleId);
```
schmidt-sebastian, line 187: s/`@Nullable`/`@NotNull`

schmidt-sebastian (approved):
LGTM. Sorry I must have looked at a different commit.

BenWhitehead (approved):

markarndt (commented):
@wu-hui .
I think the Node version has a clearer definition of bundleID: "The id of the bundle. When loaded on clients, client SDKs use this id and the timestamp associated with the built bundle to tell if it has been loaded already. If not specified, a random identifier will be used."

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -54,10 +54,15 @@
...
     }
 
+    /** Returns the ID for this bundle. */
+    public String getId() {
+      return this.id;
+    }
+
     /**
      * Adds a Firestore document snapshot to the bundle. Both the documents data and the document
      * read time will be included in the bundle.
```
markarndt, line 68: Remove plural: "Both the document data and...."

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
```diff
@@ -54,10 +54,15 @@
...
       this.id = id;
     }
 
+    /** Returns the ID for this bundle. */
+    public String getId() {
+      return this.id;
+    }
+
     /**
      * Adds a Firestore document snapshot to the bundle. Both the documents data and the document
```
markarndt, line 67: Remove plural "Both the document data and...."

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
markarndt, line 68: Line 124-129 below, for add() doc, remove plural so "Adds a Firestore query snapshot to..."

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
wu-hui: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreBundle.java
wu-hui: Done.

wu-hui (commented):
Thanks Mark, I updated the comment for bundle ID.

PR #478, opened 2020-11-25
feat: add CollectionGroup#getPartitions(long)

BenWhitehead:
Add new method to `CollectionGroup` allowing `getPartitions` to return a `Future<List<QueryPartition>>`. This new method behaves similar to the `CollectionGroup#getPartitions(long, ApiStreamObserver<QueryPartition>)` providing a Future centric api.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
@@ -73,7 +73,7 @@ public class CollectionGroup extends Query {
       // Short circuit if the user only requested a single partition.
       observer.onNext(new QueryPartition(partitionQuery, null, null));
     } else {
-      PartitionQueryRequest.Builder request = buildRequest(desiredPartitionCount);
+      PartitionQueryRequest request = buildRequest(desiredPartitionCount);
 
       final FirestoreClient.PartitionQueryPagedResponse response;
       final TraceUtil traceUtil = TraceUtil.getInstance();
@@ -82,7 +82,7 @@ public class CollectionGroup extends Query {
         response =
             ApiExceptions.callAndTranslateApiException(
                 rpcContext.sendRequest(
-                    request.build(), rpcContext.getClient().partitionQueryPagedCallable()));
+                    request, rpcContext.getClient().partitionQueryPagedCallable()));
 
         consumePartitions(
             response,
@@ -110,14 +110,14 @@ public class CollectionGroup extends Query {
       return ApiFutures.immediateFuture(
           Collections.singletonList(new QueryPartition(partitionQuery, null, null)));
     } else {
-      PartitionQueryRequest.Builder request = buildRequest(desiredPartitionCount);
+      PartitionQueryRequest request = buildRequest(desiredPartitionCount);
 
       final TraceUtil traceUtil = TraceUtil.getInstance();
       Span span = traceUtil.startSpan(TraceUtil.SPAN_NAME_PARTITIONQUERY);
       try (Scope scope = traceUtil.getTracer().withSpan(span)) {
         return ApiFutures.transform(
             rpcContext.sendRequest(
-                request.build(), rpcContext.getClient().partitionQueryPagedCallable()),
+                request, rpcContext.getClient().partitionQueryPagedCallable()),
             new ApiFunction<PartitionQueryPagedResponse, List<QueryPartition>>() {
               @Override
               public List<QueryPartition> apply(PartitionQueryPagedResponse response) {
@@ -144,7 +144,7 @@ public class CollectionGroup extends Query {
     }
   }
 
-  private PartitionQueryRequest.Builder buildRequest(long desiredPartitionCount) {
+  private PartitionQueryRequest buildRequest(long desiredPartitionCount) {
     Preconditions.checkArgument(
         desiredPartitionCount > 0, "Desired partition count must be one or greater");
 
@@ -155,7 +155,7 @@ public class CollectionGroup extends Query {
     // Since we are always returning an extra partition (with en empty endBefore cursor), we
     // reduce the desired partition count by one.
     request.setPartitionCount(desiredPartitionCount - 1);
-    return request;
+    return request.build();
   }
 
   private void consumePartitions(
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
@@ -16,16 +16,25 @@
 
 package com.google.cloud.firestore;
 
+import com.google.api.core.ApiFunction;
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutures;
 import com.google.api.gax.rpc.ApiException;
 import com.google.api.gax.rpc.ApiExceptions;
 import com.google.api.gax.rpc.ApiStreamObserver;
 import com.google.cloud.firestore.v1.FirestoreClient;
+import com.google.cloud.firestore.v1.FirestoreClient.PartitionQueryPagedResponse;
+import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.util.concurrent.MoreExecutors;
 import com.google.firestore.v1.Cursor;
 import com.google.firestore.v1.PartitionQueryRequest;
 import io.opencensus.common.Scope;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Status;
+import java.util.Collections;
+import java.util.List;
 import javax.annotation.Nullable;
 
 /**
@@ -33,6 +42,9 @@ import javax.annotation.Nullable;
  * subcollection with a specific collection ID.
  */
 public class CollectionGroup extends Query {
+
+  final Query partitionQuery;
+
   CollectionGroup(FirestoreRpcContext<?> rpcContext, String collectionId) {
     super(
         rpcContext,
@@ -41,6 +53,9 @@ public class CollectionGroup extends Query {
             .setCollectionId(collectionId)
             .setAllDescendants(true)
             .build());
+
+    // Partition queries require explicit ordering by __name__.
+    partitionQuery = orderBy(FieldPath.DOCUMENT_ID);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


BenWhitehead:
With the code that is in the PR right now, the future transform takes place on one of the gax threads leaving the invoking thread unblocked as can be seen in the following log snippet.

```
2021-01-07T21:38:31.080 [main] com.google.cloud.firestore.CollectionGroup - getPartitions(3)...
2021-01-07T21:38:31.098 [main] com.google.cloud.firestore.CollectionGroup - getPartitions(3) complete
2021-01-07T21:38:31.098 [main] com.google.cloud.firestore.CollectionGroup - calling future.get()...
2021-01-07T21:38:31.255 [Gax-4] com.google.cloud.firestore.CollectionGroup - transform->apply()
2021-01-07T21:38:31.260 [main] com.google.cloud.firestore.CollectionGroup - calling future.get() complete
```

If I change the implementation to adapt the existing ApiStreamObserver method like so
```
  public ApiFuture<List<QueryPartition>> getPartitionss(long desiredPartitionCount) {
    final SettableApiFuture<List<QueryPartition>> future = SettableApiFuture.create();
    getPartitions(desiredPartitionCount, new ApiStreamObserver<QueryPartition>() {
      final List<QueryPartition> accumulator = new ArrayList<>();
      @Override
      public void onNext(QueryPartition value) {
        accumulator.add(value);
      }

      @Override
      public void onError(Throwable t) {
        future.setException(t);
      }

      @Override
      public void onCompleted() {
        future.set(accumulator);
      }
    });
    return future;
  }
```
The invocations on the observer are done from the invoking thread meaning the future isn't really effective.

```
2021-01-07T21:38:50.196 [main] com.google.cloud.firestore.CollectionGroup - getPartitions(3)...
2021-01-07T21:38:50.301 [main] com.google.cloud.firestore.CollectionGroup - CollectionGroup.onNext
2021-01-07T21:38:50.303 [main] com.google.cloud.firestore.CollectionGroup - CollectionGroup.onNext
2021-01-07T21:38:50.303 [main] com.google.cloud.firestore.CollectionGroup - CollectionGroup.onCompleted
2021-01-07T21:38:50.303 [main] com.google.cloud.firestore.CollectionGroup - getPartitions(3) complete
2021-01-07T21:38:50.304 [main] com.google.cloud.firestore.CollectionGroup - calling future.get()...
2021-01-07T21:38:50.304 [main] com.google.cloud.firestore.CollectionGroup - calling future.get() complete
```

I thought about using `rpcContext.getClient().getExecutor().submit(...)` to wrap the call to `getPartitions` but that executor is the `grpc-transport` threadpool which we definitely don't want to block.
```
2021-01-07T21:44:20.290 [main] com.google.cloud.firestore.CollectionGroup - getPartitions(3)...
2021-01-07T21:44:20.292 [main] com.google.cloud.firestore.CollectionGroup - getPartitions(3) complete
2021-01-07T21:44:20.293 [main] com.google.cloud.firestore.CollectionGroup - calling future.get()...
2021-01-07T21:44:20.453 [grpc-transport-0] com.google.cloud.firestore.CollectionGroup - CollectionGroup.onNext
2021-01-07T21:44:20.455 [grpc-transport-0] com.google.cloud.firestore.CollectionGroup - CollectionGroup.onNext
2021-01-07T21:44:20.455 [grpc-transport-0] com.google.cloud.firestore.CollectionGroup - CollectionGroup.onCompleted
2021-01-07T21:44:20.455 [main] com.google.cloud.firestore.CollectionGroup - calling future.get() complete
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
```diff
@@ -101,4 +109,60 @@ public void getPartitions(
 
     observer.onCompleted();
   }
+
+  public ApiFuture<List<QueryPartition>> getPartitions(long desiredPartitionCount) {
```
schmidt-sebastian, line 113: I wonder if you we could do something like this: https://github.com/googleapis/java-firestore/blob/edaa5395be0353fb261d954429c624623bc4e346/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java#L383 and expose an Iterable of QueryPartitions? I assume this won't work because of some threading issues, but it would be nice to re-use an existing pattern.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
```diff
@@ -101,4 +109,60 @@ public void getPartitions(
...
+    } else {
+      PartitionQueryRequest.Builder request = PartitionQueryRequest.newBuilder();
+      request.setStructuredQuery(queryWithDefaultOrder.buildQuery());
+      request.setParent(options.getParentPath().toString());
+
+      // Since we are always returning an extra partition (with en empty endBefore cursor), we
+      // reduce the desired partition count by one.
+      request.setPartitionCount(desiredPartitionCount - 1);
+
+      final TraceUtil traceUtil = TraceUtil.getInstance();
```
schmidt-sebastian, line 133: All the lines above here could be factored out into a helper and shared.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
```diff
@@ -101,4 +109,60 @@ public void getPartitions(
...
+      try (Scope scope = traceUtil.getTracer().withSpan(span)) {
+        return ApiFutures.transform(
+            rpcContext.sendRequest(
+                request.build(), rpcContext.getClient().partitionQueryPagedCallable()),
+            new ApiFunction<PartitionQueryPagedResponse, List<QueryPartition>>() {
+              @Override
+              public List<QueryPartition> apply(PartitionQueryPagedResponse input) {
+                ImmutableList.Builder<QueryPartition> partitions = ImmutableList.builder();
+                Object[] lastCursor = null;
+                for (Cursor cursor : input.iterateAll()) {
```
schmidt-sebastian, line 144: I think with some trickery these lines here could also be factored out into a common function. Optional.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
```diff
@@ -101,4 +109,60 @@ public void getPartitions(
...
+                return partitions.build();
+              }
+            },
+            MoreExecutors.directExecutor());
+      } catch (ApiException exception) {
+        span.setStatus(Status.UNKNOWN.withDescription(exception.getMessage()));
+        throw FirestoreException.apiException(exception);
+      } finally {
+        span.end(TraceUtil.END_SPAN_OPTIONS);
+      }
```
schmidt-sebastian, line 165: FWIW these Codecov warnings seem incorrect and make it hard to read this code...

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
BenWhitehead: If we go this route we lose the ability to chain the result of this method into a series of queries with the following pattern:

```java
    ApiFuture<List<QuerySnapshot>> allQueries = ApiFutures.transformAsync(
        firestore.collectionGroup(randomColl.getId()).getPartitions(3),
        new ApiAsyncFunction<List<QueryPartition>, List<QuerySnapshot>>() {
          @Override
          public ApiFuture<List<QuerySnapshot>> apply(List<QueryPartition> input) {
            logit("ApiAsyncFunction.apply");
            List<ApiFuture<QuerySnapshot>> futures = new ArrayList<>();
            for (QueryPartition part : input) {
              logit("part = %s", part);
              futures.add(part.createQuery().get());
            }
            return ApiFutures.allAsList(futures);
          }
        },
        MoreExecutors.directExecutor()
    );
    List<QuerySnapshot> querySnapshots = allQueries.get();
```

```
2021-01-08T23:52:57.989  [main] com.google.cloud.firestore.CollectionGroup - transformAsync...
2021-01-08T23:52:58.008  [main] com.google.cloud.firestore.CollectionGroup - transformAsync Complete
2021-01-08T23:52:58.008  [main] com.google.cloud.firestore.CollectionGroup - allQueries.get()...
2021-01-08T23:52:58.111  [Gax-4] com.google.cloud.firestore.CollectionGroup - transform->apply()
2021-01-08T23:52:58.116  [Gax-4] com.google.cloud.firestore.CollectionGroup - ApiAsyncFunction.apply
2021-01-08T23:52:58.117  [Gax-4] com.google.cloud.firestore.CollectionGroup - part = com.google.cloud.firestore.QueryPartition@a9b0a06e
2021-01-08T23:52:58.131  [Gax-4] com.google.cloud.firestore.CollectionGroup - part = com.google.cloud.firestore.QueryPartition@19a03734
2021-01-08T23:52:58.309  [main] com.google.cloud.firestore.CollectionGroup - allQueries.get() complete
```

There is also additional state that has to be tracked because the iterator we return would have to potentially run past the end of the iterator returned by `response.iterateAll()` not to mention the iteration state now has to be split across the hasNext and next methods. (It's possible but actually much more code, so I can swap it in if you want me to).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
BenWhitehead: Agreed, not sure why they're showing up for this change. There is a test added below that is running this code.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
BenWhitehead: I could pull out up to the PartitionQueryRequest.Builder line, but further than that and I get into the territory have having to model an abstract lifecycle type object to allow the individual handling of the single case. Which also would extend to the loop below. If we had some existing "addable" or similar type we could use it to wrap the observer and the list but I don't think it's worth it here.

What do you think, should I still work to factor this out?

schmidt-sebastian: I played a bit with this (but didn't run a test): https://gist.github.com/schmidt-sebastian/ce122b16bd13837cf9c01cfe38783bf1

Let me know what you think.

BenWhitehead: I see, thanks for sharing the patch I've incorporated it into the PR.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
```diff
@@ -80,25 +83,92 @@ public void getPartitions(
...
+        throw FirestoreException.forApiException(exception);
+      } finally {
+        span.end(TraceUtil.END_SPAN_OPTIONS);
       }
-      observer.onNext(new QueryPartition(queryWithDefaultOrder, lastCursor, null));
     }
+  }
 
-    observer.onCompleted();
+  private PartitionQueryRequest.Builder buildRequest(long desiredPartitionCount) {
```
schmidt-sebastian, line 147: I think this could just return PartitionQueryRequest. We don't modify this builder anymore. Sorry for not noticing earlier.

schmidt-sebastian (approved):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
BenWhitehead: No worries, I'll clean it up before merging.

PR #497, opened 2021-01-14
fix: add @BetaApi, make BulkWriter public, and refactor Executor

thebrianchen:
Add `@BetaApi` annotation to BulkWriter, along with making the `BulkWriterOptions` class public. Also refactored the executor into the options object, to better mirror the `Transaction` API.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
@@ -17,6 +17,7 @@
 package com.google.cloud.firestore;
 
 import com.google.api.core.ApiFuture;
+import com.google.api.core.BetaApi;
 import com.google.api.core.InternalExtensionOnly;
 import com.google.api.gax.rpc.ApiStreamObserver;
 import com.google.cloud.Service;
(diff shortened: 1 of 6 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
```diff
@@ -168,9 +169,29 @@ void getAll(
...
+   * @see <a href=https://cloud.google.com/datastore/docs/best-practices#ramping_up_traffic>Ramping
+   *     up traffic</a>
+   */
+  @BetaApi
   @Nonnull
   BulkWriter bulkWriter();
 
+  /**
+   * Creates a {@link BulkWriter} instance, used for performing multiple writes in parallel.
+   * Gradually ramps up writes as specified by the 500/50/5 rule.
```
schmidt-sebastian, line 185: This isn't necessarily true for all `options` object, is it?

schmidt-sebastian (approved):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Firestore.java
thebrianchen: Added "unless otherwise configured by a BulkWriterOptions object".

PR #508, opened 2021-01-24
Removed Jackson dependency

### LINKED ISSUE #506: Remove dependency on jackson2
Per go/json-madness

abhijeetshuklaoist:
Signed-off-by: Abhijeet Shukla <abhijeetshuklaoist@gmail.com>

Fixes #506 ☕️

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
@@ -19,8 +19,6 @@ package com.google.cloud.firestore;
 import static org.junit.Assert.assertEquals;
 import static org.mockito.Mockito.doAnswer;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
 import com.google.api.gax.retrying.RetrySettings;
@@ -55,13 +53,15 @@ import com.google.firestore.v1.StructuredQuery.FieldFilter;
 import com.google.firestore.v1.StructuredQuery.UnaryFilter;
 import com.google.firestore.v1.Value;
 import com.google.firestore.v1.Write;
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
 import com.google.protobuf.ByteString;
 import com.google.protobuf.Empty;
 import com.google.protobuf.GeneratedMessageV3;
 import com.google.protobuf.Message;
 import com.google.protobuf.NullValue;
 import com.google.type.LatLng;
-import java.io.IOException;
+import java.lang.reflect.Type;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
@@ -970,12 +970,9 @@ public final class LocalFirestoreHelper {
   }
 
   private static Map<String, Object> fromJsonString(String json) {
-    try {
-      ObjectMapper mapper = new ObjectMapper();
-      return mapper.readValue(json, new TypeReference<Map<String, Object>>() {});
-    } catch (IOException e) {
-      throw new RuntimeException(e);
-    }
+    Type type = new TypeToken<Map<String, Object>>() {}.getType();
+    Gson gson = new Gson();
+    return gson.fromJson(json, type);
   }
 
   public static Map<String, Object> fromSingleQuotedString(String json) {
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/MapperTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/MapperTest.java
@@ -20,9 +20,11 @@ import static com.google.cloud.firestore.LocalFirestoreHelper.fromSingleQuotedSt
 import static com.google.cloud.firestore.LocalFirestoreHelper.mapAnyType;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import com.google.cloud.Timestamp;
 import com.google.cloud.firestore.annotation.DocumentId;
 import com.google.cloud.firestore.annotation.Exclude;
 import com.google.cloud.firestore.annotation.PropertyName;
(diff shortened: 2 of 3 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -970,18 +978,88 @@ public static String autoId() {
...
+    gsonBuilder.registerTypeAdapterFactory(CustomObjectTypeAdapter.FACTORY);
+    Gson gson = gsonBuilder.create();
+    return gson.fromJson(json, type);
   }
 
   public static Map<String, Object> fromSingleQuotedString(String json) {
     return fromJsonString(json.replace("'", "\""));
   }
 
+  static class CustomObjectTypeAdapter extends TypeAdapter<Object> {
```
abhijeetshuklaoist, line 992: Needed to write this Custom TypeAdapter because GSON considers every numeric value to be Double by deafult and there is no neat way around it.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/MapperTest.java
```diff
@@ -1214,7 +1214,7 @@ public void primitiveDeserializeLong() {
   @Test
   public void primitiveDeserializeWrongTypeMap() {
     assertExceptionContains(
-        "Failed to convert value of type java.util.LinkedHashMap to String "
+        "Failed to convert value of type com.google.gson.internal.LinkedTreeMap to String "
```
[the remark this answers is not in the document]
abhijeetshuklaoist: I agree with you, I have updated this particular test and removed exception message. I will refactor other similar tests in another PR.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
elharo: In javaScript and JSON all numbvers are doubles. There are no ints.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -970,18 +978,88 @@ public static String autoId() {
...
+          while (in.hasNext()) {
+            map.put(in.nextName(), read(in));
+          }
+          in.endObject();
+          return map;
+
+        case STRING:
+          return in.nextString();
+
+        case NUMBER:
```
elharo, line 1032: Can you simply use doubles here?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/MapperTest.java
```diff
@@ -1213,9 +1222,7 @@ public void primitiveDeserializeLong() {
 
   @Test
   public void primitiveDeserializeWrongTypeMap() {
-    assertExceptionContains(
-        "Failed to convert value of type java.util.LinkedHashMap to String "
-            + "(found in field 'value')",
+    assertExceptionThrows(
```
elharo, line 1225: This is overengineered. You can move the try-catch block into this method and not use a runnable.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/MapperTest.java
```diff
@@ -972,6 +972,15 @@ private static void assertExceptionContains(String partialMessage, Runnable run)
     }
   }
 
+  private static void assertExceptionThrows(Runnable run) {
+    try {
+      run.run();
+      fail("Expected exception not thrown");
+    } catch (RuntimeException ignored) {
```
elharo, line 979: This should be a more specific exception

schmidt-sebastian (changes requested):
Another approach for this PR would be to simply drop the JSON parsing from the callsites. The JSON parser is only used in the Unit tests and it seems straightforward to replace its usage. Instead of:


```
  @Test
  public void primitiveDeserializeString() {
    StringBean bean = deserialize("{'value': 'foo'}", StringBean.class);
    assertEquals("foo", bean.value);
    ...
```

We could just do:

```
  @Test
  public void primitiveDeserializeString() {
    StringBean bean = deserialize(map("value",  "foo"), StringBean.class); // 
    assertEquals("foo", bean.value);

  private static <T> T deserialize(Map<String, Object> value, Class<T> clazz) ...
```

`deserialize` could also be:

```
  private static <T> T deserialize(Class<T> clazz, String key, Object value, Object ... moreKeysAndValues) ...
```

Which would make the callsites even more concise.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
abhijeetshuklaoist: Sure, I will remove custom object adapter altogether and will update the tests accordingly because existing tests need numbers to be Integers, long and doubles

google-cloud-firestore/src/test/java/com/google/cloud/firestore/MapperTest.java
abhijeetshuklaoist: Will do

abhijeetshuklaoist:
@googlebot I fixed it.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/MapperTest.java
```diff
@@ -1211,17 +1223,9 @@ public void primitiveDeserializeLong() {
...
-    assertExceptionContains(
-        "Failed to convert value of type java.util.LinkedHashMap to String "
-            + "(found in field 'value')",
-        new Runnable() {
-          @Override
-          public void run() {
-            deserialize("{'value': {'foo': 'bar'}}", StringBean.class);
-          }
-        });
+    deserialize("{'value': {'foo': 'bar'}}", StringBean.class);
```
schmidt-sebastian, line 1228: This test loses the exception text, which verifies that we tell the developer where the exception occurred.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/MapperTest.java
```diff
@@ -2486,15 +2492,15 @@ public void run() {
...
+    assertJson("{'value': 1}", serialize(Collections.singletonMap("value", 1.0)));
   }
 
   @Test
   public void settersCanOverridePrimitiveSettersParsing() {
     NonConflictingSetterSubBean bean =
         deserialize("{'value': 2}", NonConflictingSetterSubBean.class);
     // sub-bean converts to negative value
-    assertEquals(-2, bean.value);
+    assertEquals(-2, bean.value, 0);
```
schmidt-sebastian, line 2503: What is the `0` for? Isn't the third argument meant to be a descriptive text?

schmidt-sebastian (approved):

Issue #491, opened 2020-12-28
Null field value is treated inconsistently in DocumentSnapshot get{Type} methods (return null vs throw NPE)


AndrewJudson:
Compare
https://github.com/googleapis/java-firestore/blob/master/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java#L336

with

https://github.com/googleapis/java-firestore/blob/master/google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java#L348

In the first example, if the field does not exist or is null, you return null. In the second case, you throw an NPE. This is confusing when the behavior is not consistent between different field types, and it is incorrect to label the methods that throw an NPE as Nullable, as they will never return null.

schmidt-sebastian:
This looks like a bug in toDate(). Fix is coming.

PR #513, opened 2021-01-26
fix: don't throw NPE in DocumentSnapshot.getDate() when value is missing

schmidt-sebastian:
Fixes https://github.com/googleapis/java-firestore/issues/491


thebrianchen (approved):
lgtm

PR #523, opened 2021-02-03
chore: simplify BulkWriter

schmidt-sebastian:
This is my attempt to simplify some of the callback logic in BulkWriter, and will hopefully allow us add Backoff handling for retries pretty easily.

The overall changes are:

- There are no more batch queues. Instead, there is only ever a single batch. If this batch reaches capacity, it is send.
- To allow 'flush' to block, I am instead keeping a list of pending operations.

One behavior change is that it is now possible that a retry will get scheduled after other operations in a flush. Let's say we have the calls:

- Op1
- Op2
- Flush
- Op3

And Op1 fails. The write order could now be: Op1, Op2, Op3, Op1. This breaks a single unit test, but I think it is sane behavior as it allows us to make progress while we wait for more reponses.

Port of https://github.com/googleapis/nodejs-firestore/pull/1379

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
@@ -20,7 +20,7 @@ import com.google.api.core.ApiAsyncFunction;
 import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
-import com.google.api.core.SettableApiFuture;
+import com.google.api.gax.rpc.ApiException;
 import com.google.cloud.Timestamp;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
@@ -34,40 +34,22 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
-import javax.annotation.Nullable;
+import java.util.concurrent.Executor;
 
 /** Used to represent a batch on the BatchQueue. */
 class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
-  /**
-   * Used to represent the state of batch.
-   *
-   * <p>Writes can only be added while the batch is OPEN. For a batch to be sent, the batch must be
-   * READY_TO_SEND. After a batch is sent, it is marked as SENT.
-   */
-  enum BatchState {
-    OPEN,
-    READY_TO_SEND,
-    SENT,
-  }
 
-  private BatchState state = BatchState.OPEN;
-
-  private final List<SettableApiFuture<BatchWriteResult>> pendingOperations = new ArrayList<>();
+  private final List<BulkWriterOperation> pendingOperations = new ArrayList<>();
   private final Set<DocumentReference> documents = new CopyOnWriteArraySet<>();
-  private final int maxBatchSize;
+  private final Executor executor;
 
-  BulkCommitBatch(FirestoreImpl firestore, int maxBatchSize) {
+  BulkCommitBatch(FirestoreImpl firestore, Executor executor) {
     super(firestore);
-    this.maxBatchSize = maxBatchSize;
+    this.executor = executor;
   }
 
-  @Override
-  boolean isCommitted() {
-    return state == BatchState.SENT;
-  }
-
-  ApiFuture<WriteResult> wrapResult(DocumentReference documentReference) {
-    return processLastOperation(documentReference);
+  ApiFuture<WriteResult> wrapResult(int writeIndex) {
+    return pendingOperations.get(writeIndex).getFuture();
   }
 
   /**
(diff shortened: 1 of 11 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOperation.java
```diff
@@ -0,0 +1,174 @@
...
+  private final SettableApiFuture<WriteResult> future = SettableApiFuture.create();
+  private final DocumentReference documentReference;
+  private final BulkWriter.OperationType operationType;
+  private final ApiFunction<BulkWriterOperation, Void> scheduleWriteCallback;
+  private final Executor successExecutor;
+  private final BulkWriter.WriteResultCallback successListener;
+  private final Executor errorExecutor;
+  private final BulkWriter.WriteErrorCallback errorListener;
+
+  private int failedAttempts;
```
thebrianchen, line 40: possible nit: Do we initialize variables in java even if it's implicit?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOperation.java
```diff
@@ -0,0 +1,174 @@
...
+import com.google.api.core.ApiFutures;
+import com.google.api.core.SettableApiFuture;
+import com.google.common.util.concurrent.MoreExecutors;
+import java.util.concurrent.Executor;
+
+/**
+ * Represents a single write for BulkWriter, encapsulating operation dispatch and error handling.
+ */
+class BulkWriterOperation {
+  private final SettableApiFuture<WriteResult> future = SettableApiFuture.create();
```
thebrianchen, line 31: optional: rename to `operationFuture` to avoid confusion with the local future variables in the error and success handlers (or maybe it's just me that gets confused).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriterOperation.java
```diff
@@ -0,0 +1,174 @@
...
+            failedAttempts);
+
+    final SettableApiFuture<Void> future = SettableApiFuture.create();
+
+    ApiFutures.addCallback(
+        invokeUserErrorCallback(bulkWriterException),
+        new ApiFutureCallback<Boolean>() {
+          @Override
+          public void onFailure(Throwable throwable) {
+            BulkWriterOperation.this.future.setException(throwable);
```
thebrianchen, line 99: If you rename the instance `future` variable, can you remove the need to use `BulkWriterOperation.this`?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -73,7 +74,21 @@
...
+              false));
+
+  private static final ApiFuture<GeneratedMessageV3> RETRYABLE_FAILED_FUTURE =
+      ApiFutures.immediateFailedFuture(
+          new ApiException(
+              new IllegalStateException("Mock batchWrite failed in test"),
+              GrpcStatusCode.of(Status.Code.ABORTED),
+              true));
+
+  // @Rule public Timeout timeout = new Timeout(500, TimeUnit.MILLISECONDS);
```
thebrianchen, line 91: Can't have BulkWriter mysteriously flaking everywhere again.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -73,7 +74,21 @@
 @RunWith(MockitoJUnitRunner.class)
 public class BulkWriterTest {
 
-  @Rule public Timeout timeout = new Timeout(500, TimeUnit.MILLISECONDS);
+  private static final ApiFuture<GeneratedMessageV3> FAILED_FUTURE =
```
thebrianchen, line 77: This is a very sleek cleanup.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
```diff
@@ -34,159 +34,108 @@
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
-import javax.annotation.Nullable;
+import java.util.concurrent.Executor;
 
 /** Used to represent a batch on the BatchQueue. */
```
thebrianchen, line 39: Update class name?
My attempt: `A batch that holds and schedules BulkWriter operations.`

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
```diff
@@ -34,159 +34,108 @@
...
-  }
-
-  boolean isOpen() {
-    return state == BatchState.OPEN;
-  }
-
-  boolean isReadyToSend() {
-    return state == BatchState.READY_TO_SEND;
+  void enqueueOperation(BulkWriterOperation operation) {
+    boolean added = documents.add(operation.getDocumentReference());
```
thebrianchen, line 136: Wow, that optimization to add then check 🤯

thebrianchen (changes requested):
Thanks for porting this! Unfortunately, the unit tests flake when you run them until failure on IntelliJ. I suspect this might be because my old race conditions were covered by using `MoreExecutors.directExecutor()` instead of calling the executor by name, which forced the running thread to continue in the same context.

schmidt-sebastian (commented):
Thanks for pushing me to fix the flakes. I think this uncovered an actual issue where the batch was sometimes modified by multiple threads (through the backoff handler as well as the user input).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
thebrianchen: The main issue I had was that "BatchQueue" is no longer a term used in BulkWriter.

thebrianchen (approved):
Thanks for diving into this one, really appreciate it :)

PR #529, opened 2021-02-06
chore: fix some potential threading issues in BulkWriter

schmidt-sebastian:
- lastOperation is read from multiple threads
- the rate limiter is not thread-safe

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
@@ -157,12 +157,11 @@ public final class BulkWriter implements AutoCloseable {
   private boolean closed = false;
 
   /** Rate limiter used to throttle requests as per the 500/50/5 rule. */
+  @GuardedBy("lock")
   private final RateLimiter rateLimiter;
 
   private WriteResultCallback successListener = DEFAULT_SUCCESS_LISTENER;
-
   private WriteErrorCallback errorListener = DEFAULT_ERROR_LISTENER;
-
   private Executor successExecutor;
   private Executor errorExecutor;
 
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
@@ -73,6 +73,7 @@ public final class BulkWriter implements AutoCloseable {
     UPDATE,
     DELETE
   }
+
   /** The maximum number of writes that can be in a single batch. */
   public static final int MAX_BATCH_SIZE = 20;
 
@@ -150,7 +151,7 @@ public final class BulkWriter implements AutoCloseable {
    * A pointer to the tail of all active BulkWriter applications. This pointer is advanced every
    * time a new write is enqueued.
    */
-  private ApiFuture<Void> lastOperation = ApiFutures.immediateFuture(null);
+  private volatile ApiFuture<Void> lastOperation = ApiFutures.immediateFuture(null);
 
   /** Whether this BulkWriter instance is closed. Once closed, it cannot be opened again. */
   private boolean closed = false;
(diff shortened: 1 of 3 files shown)


schmidt-sebastian:
Hm, this is now super flaky.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -675,7 +675,7 @@ public boolean onError(BulkWriterException error) {
   }
 
   @Test
-  public void retriesMaintainCorrectWriteResolutionOrdering() throws Exception {
+  public void retryResolvesBeforeFlush() throws Exception {
```
schmidt-sebastian, line 678: I don't think I will be able to deflake the original test as the order of whether doc1, doc2, doc1 or doc1, doc1, doc2 is enqueued is no longer deterministic. I don't think that's a problem though, but it breaks this test.

thebrianchen: This behavior is still reasonable though, so it should be fine.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -150,18 +151,17 @@ public boolean onError(BulkWriterException error) {
    * A pointer to the tail of all active BulkWriter applications. This pointer is advanced every
    * time a new write is enqueued.
    */
-  private ApiFuture<Void> lastOperation = ApiFutures.immediateFuture(null);
+  private volatile ApiFuture<Void> lastOperation = ApiFutures.immediateFuture(null);
```
thebrianchen, line 154: Woah, did not know about `volatile` in Java.

If `lastOperation` is only accessed inside the `synchronized` call, why do we still need the `volatile` label? My stackoverflow hunting initially led me to think of volatile variables as a simpler/weaker form of synchronization than `synchronized`, which makes `volatile` redundant ([link](https://stackoverflow.com/questions/22862226/java-is-volatile-required-with-synchronized), [another link](https://stackoverflow.com/questions/32974336/is-volatile-needed-in-case-of-only-synchronized-access)). However, after reading another conflicting [post](https://stackoverflow.com/questions/50298456/redundant-volatile-in-cheap-read-write-lock) that argues `volatile` is required for reading the variable, I'm now uncertain.

thebrianchen (approved):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: Actually, I realize we're reading `lastOperation` in `flush()`, which is out of the `synchronized` block, which then means that we need to use `volatile`.

dconeybe: Drive by comment... Although I can't find the documentation to back this up, I'm 99% sure that you basically *never* need to declare a variable as `volatile` in Java. In my entire career I have never needed to use `volatile`, even in heavily multi-threaded code. I think that the only cases where you would need to use `volatile` would be (a) if you're implementing some sort of custom lock-free synchronization on the variable (e.g. busy waiting) or (b) if the variable's value could be changed asynchronously from native code.

schmidt-sebastian: `volatile` ensures that the value that is read is never read from the CPU cache. Without this, `flush()` is not guaranteed to see the latest state of `lastOperation` if `lastOperation` was modified by another thread.

cc @wilhuff

wilhuff: This `volatile` is an indicator of broader problems within this implementation.

  * `verifyNotClosed` is called unlocked and accesses `closed` without locking
  * `writesEnqueued` is written to by `executeWrite` also without locking

In Java, `volatile` is expensive relative to the underlying atomic operations that CPUs can perform--it's like a half lock. If you made all of these values `volatile` it would actually be more expensive than just synchronizing. In Java synchronization is very cheap in the uncontended case, and this doesn't seem to be a class that should have to optimize for that.

My advice:
  * don't use volatile; just use synchronization consistently
  * acquire the lock nearer the API boundary so you're not acquiring it multiple times per private method used to implement the API call
  * name your internal methods to show you assume the lock is held (e.g. `verifyNotClosedLocked`)
  * avoid trying to be fancy with multithreading at all

wilhuff: That is to say, at a high level, the way to fix the problem with `flush` accessing `lastOperation` outside a `synchronized` block is to add a `synchronized` block. You should also fix the other unsynchonized accesses by adding `synchronized` blocks. Once you do, you'll notice you're synchronizing multiple times per API invocation, in which case you can hoist the synchronized blocks.

Also, see go/atomic-danger internally.

schmidt-sebastian: Thanks for the answer, very helpful. I will look at this a bit more tomorrow (and will likely add more synchronization).

To "defend" the current state of the PR: Both `this.closed` and `this.writesEnqueued` is only accessed from the user thread. `bulkCommitBatch`, `lastOperation` and `rateLimiter` are accessed both from the user thread and as a result of a BatchCommit response, in which case they are accessed from the BulkWriter thread.

Time to read go/atomic-danger

schmidt-sebastian: I updated this PR to use synchronization. PTAL.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -171,12 +177,6 @@ public boolean onError(BulkWriterException error) {
    */
   private boolean writesEnqueued = false;
```
thebrianchen, line 178: Since all accesses to writes enqueued are under lock, should we add a `@GuardedBy` annotation?

thebrianchen (approved):
I see how moving the synchronized calls up and using the `locked` method naming makes the code easier to understand, thanks!

Also, why were the `@GuardedBy` annotations removed? I think having them makes it much easier to quickly see which variables are accessed under lock.

For example, I'm looking at some Android [source code](http://androidxref.com/5.0.0_r2/xref/frameworks/base/core/java/android/app/ContextImpl.java#1007), and they use the `Locked` method naming convention along with `@GuardedBy` annotations for the instance variables that are accessed under lock.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -131,35 +130,42 @@ public boolean onError(BulkWriterException error) {
...
+
+  private final FirestoreImpl firestore;
 
-  /** Lock object for `bulkCommitBatch`. */
+  // Executor used to run all BulkWriter operations. BulkWriter uses its own executor since we
+  // don't want to block a gax/grpc executor while running user error and success callbacks.
+  private final ScheduledExecutorService bulkWriterExecutor;
+
+  /**
+   * Lock object for all mutable state in bulk writer. Bulk Writer state is accessed from the user
```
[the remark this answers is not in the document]
schmidt-sebastian: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
schmidt-sebastian: I removed them since this now applies to all state. Based on the Android link, I added them back. There is also some tooling that catches violations when `GuardedBy` object is accessed under lock (i.e. in Android Studio's build), so it is likely always better to add this annotations.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
dconeybe: Wow my understanding of Java `volatile` was very mis-guided. I just read https://developer.android.com/reference/java/util/concurrent/atomic/package-summary and the documents to which it links.

I learned that the following code could loop forever if `this.done` is *not* declared `volatile`:

```java
while (!this.done) {
  Thread.sleep(100);
}
```

The compiler could choose to read `this.done` exactly once and re-use that cached value forever, even if another thread comes in and sets `this.done` to `true`. Declaring `this.done` as `volatile` would preclude this optimization and the thread would *eventually* see `this.done` set to `true` by another thread. I did not know this and have definitely coded some buggy busy waits in the past!

PR #622, opened 2021-05-05
feat: add recursiveDelete() to Firestore

thebrianchen:
Porting from node. Merging to a feature branch until it's good to release.

Several PRs are contained in here: [1](https://github.com/googleapis/nodejs-firestore/pull/1453), [2](https://github.com/googleapis/nodejs-firestore/pull/1412), [3](https://github.com/googleapis/nodejs-firestore/pull/1427).

### DIFF
--- a/google-cloud-firestore/clirr-ignored-differences.xml
+++ b/google-cloud-firestore/clirr-ignored-differences.xml
@@ -256,6 +256,6 @@
   <difference>
     <differenceType>7012</differenceType>
     <className>com/google/cloud/firestore/Firestore</className>
-    <method>com.google.cloud.firestore.Firestore recursiveDelete(*)</method>
+    <method>com.google.api.core.ApiFuture recursiveDelete(*)</method>
   </difference>
 </differences>
--- a/google-cloud-firestore/clirr-ignored-differences.xml
+++ b/google-cloud-firestore/clirr-ignored-differences.xml
@@ -252,4 +252,10 @@
     <to>*</to>
   </difference>
 
+  <!-- Recursive Delete -->
+  <difference>
+    <differenceType>7012</differenceType>
+    <className>com/google/cloud/firestore/Firestore</className>
+    <method>com.google.cloud.firestore.Firestore recursiveDelete(*)</method>
+  </difference>
 </differences>
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -135,7 +135,8 @@ class FirestoreImpl implements Firestore, FirestoreRpcContext<FirestoreImpl> {
   }
 
   @Nonnull
-  public ApiFuture<Void> recursiveDelete(DocumentReference reference, @Nonnull BulkWriter bulkWriter) {
+  public ApiFuture<Void> recursiveDelete(
+      DocumentReference reference, @Nonnull BulkWriter bulkWriter) {
     return recursiveDelete(reference, /* collectionReference= */ null, bulkWriter);
   }
 
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -120,36 +120,30 @@ class FirestoreImpl implements Firestore, FirestoreRpcContext<FirestoreImpl> {
   @Nonnull
   public ApiFuture<Void> recursiveDelete(CollectionReference reference) {
     BulkWriter writer = getBulkWriter();
-    return recursiveDelete(/* documentReference= */ null, reference, writer);
+    return recursiveDelete(reference.getResourcePath(), writer);
   }
 
   @Nonnull
   public ApiFuture<Void> recursiveDelete(CollectionReference reference, BulkWriter bulkWriter) {
-    return recursiveDelete(/* documentReference= */ null, reference, bulkWriter);
+    return recursiveDelete(reference.getResourcePath(), bulkWriter);
   }
 
   @Nonnull
   public ApiFuture<Void> recursiveDelete(DocumentReference reference) {
     BulkWriter writer = getBulkWriter();
-    return recursiveDelete(reference, /* collectionReference= */ null, writer);
+    return recursiveDelete(reference.getResourcePath(), writer);
   }
 
   @Nonnull
   public ApiFuture<Void> recursiveDelete(
       DocumentReference reference, @Nonnull BulkWriter bulkWriter) {
-    return recursiveDelete(reference, /* collectionReference= */ null, bulkWriter);
+    return recursiveDelete(reference.getResourcePath(), bulkWriter);
   }
 
-  ApiFuture<Void> recursiveDelete(
-      @Nullable DocumentReference documentReference,
-      @Nullable CollectionReference collectionReference,
-      @Nonnull BulkWriter bulkWriter) {
+  @Nonnull
+  public ApiFuture<Void> recursiveDelete(ResourcePath path, BulkWriter bulkWriter) {
     return recursiveDelete(
-        documentReference,
-        collectionReference,
-        bulkWriter,
-        RecursiveDelete.MAX_PENDING_OPS,
-        RecursiveDelete.MIN_PENDING_OPS);
+        path, bulkWriter, RecursiveDelete.MAX_PENDING_OPS, RecursiveDelete.MIN_PENDING_OPS);
   }
 
   /**
@@ -159,33 +153,8 @@ class FirestoreImpl implements Firestore, FirestoreRpcContext<FirestoreImpl> {
   @Nonnull
   @VisibleForTesting
   ApiFuture<Void> recursiveDelete(
-      @Nullable DocumentReference documentReference,
-      @Nullable CollectionReference collectionReference,
-      @Nonnull BulkWriter bulkWriter,
-      int maxLimit,
-      int minLimit) {
-    Preconditions.checkState(
-        (documentReference == null && collectionReference != null)
-            || (documentReference != null && collectionReference == null),
-        "Either documentReference or collectionReference should be set.");
-
-    ResourcePath parentPath;
-    String collectionId;
-    if (documentReference != null) {
-      // The parent is the closest ancestor document to the location we're deleting. Since we are
-      // deleting a document, the parent is the path of that document.
-      parentPath = documentReference.getResourcePath();
-      collectionId = documentReference.getParent().getId();
-    } else {
-      // The parent is the closest ancestor document to the location we're deleting. Since we are
-      // deleting a collection, the parent is the path of the document containing that collection
-      // (or the database root, if it is a root collection).
-      parentPath = collectionReference.getResourcePath().popLast();
-      collectionId = collectionReference.getId();
-    }
-    RecursiveDelete deleter =
-        new RecursiveDelete(
-            this, bulkWriter, parentPath, collectionId, documentReference, maxLimit, minLimit);
+      ResourcePath path, @Nonnull BulkWriter bulkWriter, int maxLimit, int minLimit) {
+    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, path, maxLimit, minLimit);
     return deleter.run();
   }
 
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -149,11 +149,9 @@ class FirestoreImpl implements Firestore, FirestoreRpcContext<FirestoreImpl> {
    */
   @Nonnull
   @VisibleForTesting
-  public ApiFuture<Void> recursiveDelete(
+  ApiFuture<Void> recursiveDelete(
       CollectionReference reference, BulkWriter bulkWriter, int maxLimit, int minLimit) {
-    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, reference);
-    deleter.setMaxPendingOps(maxLimit);
-    deleter.setMinPendingOps(minLimit);
+    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, reference, maxLimit, minLimit);
     return deleter.run();
   }
 
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -24,6 +24,7 @@ import com.google.api.gax.rpc.ServerStreamingCallable;
 import com.google.api.gax.rpc.UnaryCallable;
 import com.google.cloud.Timestamp;
 import com.google.cloud.firestore.spi.v1.FirestoreRpc;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.firestore.v1.BatchGetDocumentsRequest;
@@ -59,6 +60,12 @@ class FirestoreImpl implements Firestore, FirestoreRpcContext<FirestoreImpl> {
   private final FirestoreOptions firestoreOptions;
   private final ResourcePath databasePath;
 
+  /**
+   * A lazy-loaded BulkWriter instance to be used with recursiveDelete() if no BulkWriter instance
+   * is provided.
+   */
+  @Nullable private BulkWriter bulkWriterInstance;
+
   private boolean closed;
 
   FirestoreImpl(FirestoreOptions options) {
@@ -74,6 +81,14 @@ class FirestoreImpl implements Firestore, FirestoreRpcContext<FirestoreImpl> {
             + "Please explicitly set your Project ID in FirestoreOptions.");
     this.databasePath =
         ResourcePath.create(DatabaseRootName.of(options.getProjectId(), options.getDatabaseId()));
+  }
+
+  /** Lazy-load the Firestore's default BulkWriter. */
+  private BulkWriter getBulkWriter() {
+    if (bulkWriterInstance == null) {
+      bulkWriterInstance = bulkWriter();
+    }
+    return bulkWriterInstance;
   }
 
   /** Creates a pseudo-random 20-character ID that can be used for Firestore documents. */
@@ -100,6 +115,46 @@ class FirestoreImpl implements Firestore, FirestoreRpcContext<FirestoreImpl> {
   @Nonnull
   public BulkWriter bulkWriter(BulkWriterOptions options) {
     return new BulkWriter(this, options);
+  }
+
+  @Nonnull
+  public ApiFuture<Void> recursiveDelete(CollectionReference reference) {
+    BulkWriter writer = getBulkWriter();
+    RecursiveDelete deleter = new RecursiveDelete(this, writer, reference);
+    return deleter.run();
+  }
+
+  @Nonnull
+  public ApiFuture<Void> recursiveDelete(CollectionReference reference, BulkWriter bulkWriter) {
+    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, reference);
+    return deleter.run();
+  }
+
+  @Nonnull
+  public ApiFuture<Void> recursiveDelete(DocumentReference reference) {
+    BulkWriter writer = getBulkWriter();
+    RecursiveDelete deleter = new RecursiveDelete(this, writer, reference);
+    return deleter.run();
+  }
+
+  @Nonnull
+  public ApiFuture<Void> recursiveDelete(DocumentReference reference, BulkWriter bulkWriter) {
+    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, reference);
+    return deleter.run();
+  }
+
+  /**
+   * This overload is only used to test the query resumption with startAfter() once the
+   * RecursiveDelete instance has MAX_PENDING_OPS pending.
+   */
+  @Nonnull
+  @VisibleForTesting
+  public ApiFuture<Void> recursiveDelete(
+      CollectionReference reference, BulkWriter bulkWriter, int maxLimit, int minLimit) {
+    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, reference);
+    deleter.setMaxPendingOps(maxLimit);
+    deleter.setMinPendingOps(minLimit);
+    return deleter.run();
   }
 
   @Nonnull
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -120,38 +120,71 @@ class FirestoreImpl implements Firestore, FirestoreRpcContext<FirestoreImpl> {
   @Nonnull
   public ApiFuture<Void> recursiveDelete(CollectionReference reference) {
     BulkWriter writer = getBulkWriter();
-    RecursiveDelete deleter = new RecursiveDelete(this, writer, reference);
-    return deleter.run();
+    return recursiveDelete(/* documentReference= */ null, reference, writer);
   }
 
   @Nonnull
   public ApiFuture<Void> recursiveDelete(CollectionReference reference, BulkWriter bulkWriter) {
-    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, reference);
-    return deleter.run();
+    return recursiveDelete(/* documentReference= */ null, reference, bulkWriter);
   }
 
   @Nonnull
   public ApiFuture<Void> recursiveDelete(DocumentReference reference) {
     BulkWriter writer = getBulkWriter();
-    RecursiveDelete deleter = new RecursiveDelete(this, writer, reference);
-    return deleter.run();
+    return recursiveDelete(reference, /* collectionReference= */ null, writer);
   }
 
   @Nonnull
-  public ApiFuture<Void> recursiveDelete(DocumentReference reference, BulkWriter bulkWriter) {
-    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, reference);
-    return deleter.run();
+  public ApiFuture<Void> recursiveDelete(DocumentReference reference, @Nonnull BulkWriter bulkWriter) {
+    return recursiveDelete(reference, /* collectionReference= */ null, bulkWriter);
+  }
+
+  ApiFuture<Void> recursiveDelete(
+      @Nullable DocumentReference documentReference,
+      @Nullable CollectionReference collectionReference,
+      @Nonnull BulkWriter bulkWriter) {
+    return recursiveDelete(
+        documentReference,
+        collectionReference,
+        bulkWriter,
+        RecursiveDelete.MAX_PENDING_OPS,
+        RecursiveDelete.MIN_PENDING_OPS);
   }
 
   /**
-   * This overload is only used to test the query resumption with startAfter() once the
+   * This overload is not private in order to test the query resumption with startAfter() once the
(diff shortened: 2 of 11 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+    this.writer = writer;
+    this.documentReference = reference;
+    this.collectionReference = null;
+  }
+
+  public ApiFuture<Void> run() {
+    Preconditions.checkState(
+        documentsPending, "The recursive delete operation has already been completed");
+
+    // TODO: figure out if this is safe, or a way to lock this.
```
thebrianchen, line 133: Do we have to expose the lock in BulkWriter and synchronize it here?

google-cloud-firestore/clirr-ignored-differences.xml
```diff
@@ -252,4 +252,10 @@
     <to>*</to>
   </difference>
 
+  <!-- Recursive Delete -->
```
thebrianchen, line 255: I copied what I did for BulkWriter, but clirr is still unhappy. Not sure what I'm missing here...

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -102,6 +117,46 @@ public BulkWriter bulkWriter(BulkWriterOptions options) {
...
+    return deleter.run();
+  }
+
+  /**
+   * This overload is only used to test the query resumption with startAfter() once the
+   * RecursiveDelete instance has MAX_PENDING_OPS pending.
+   */
+  @Nonnull
+  @VisibleForTesting
+  public ApiFuture<Void> recursiveDelete(
```
schmidt-sebastian, line 152: This does not need to be public.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -255,13 +255,24 @@ public boolean equals(Object o) {
 
     abstract ImmutableList<FieldReference> getFieldProjections();
 
+    // Whether to select all documents under `parentPath`. By default, only
+    // collections that match `collectionId` are selected.
+    abstract boolean getKindless();
```
schmidt-sebastian, line 260: https://www.youtube.com/watch?v=AylaVaBlgCo

schmidt-sebastian: This should be `isKindless`: https://stackoverflow.com/questions/5322648/for-a-boolean-field-what-is-the-naming-convention-for-its-getter-setter

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -255,13 +255,24 @@ public boolean equals(Object o) {
...
+
+    // Whether to require consistent documents when restarting the query. By
+    // default, restarting the query uses the readTime offset of the original
+    // query to provide consistent results.
+    abstract boolean getRequireConsistency();
```
schmidt-sebastian, line 265: This should technically also not be `get`, but `is` doesn't work so you may want to keep it. I also wonder whether it should be "requires".

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1237,7 +1252,12 @@ BundledQuery toBundledQuery() {
   private StructuredQuery.Builder buildWithoutClientTranslation() {
     StructuredQuery.Builder structuredQuery = StructuredQuery.newBuilder();
     CollectionSelector.Builder collectionSelector = CollectionSelector.newBuilder();
-    collectionSelector.setCollectionId(options.getCollectionId());
+
+    // Kindless queries select all descendant documents, so we remove the collectionId field.
```
schmidt-sebastian, line 1256:
```suggestion
    // Kindless queries select all descendant documents, so we don't add the collectionId field.
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+   * Datastore allowed numeric IDs where Firestore only allows strings. Numeric IDs are exposed to
+   * Firestore as __idNUM__, so this is the lowest possible negative numeric value expressed in that
+   * format.
+   *
+   * <p>This constant is used to specify startAt/endAt values when querying for all descendants in a
+   * single collection.
+   */
+  public static final String REFERENCE_NAME_MIN_ID = "__id-9223372036854775808__";
+
+  /*!
```
schmidt-sebastian, line 51:
```suggestion
  /**
```

schmidt-sebastian: Also please rewrap comment.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+  private void incrementErrorCount(Throwable e) {
+    errorCount++;
+    lastError = e;
+  }
+
+  /**
+   * Used to test the query startAfter() resumption without needing to stream MAX_PENDING_OPS docs.
+   */
+  @VisibleForTesting
+  void setMaxPendingOps(int limit) {
```
schmidt-sebastian, line 344: Would the code be simpler if these were constructor params?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+    Preconditions.checkState(reference != null, "DocumentReference cannot be null");
+    this.firestoreRpcContext = firestoreRpcContext;
+    this.writer = writer;
+    this.documentReference = reference;
+    this.collectionReference = null;
+  }
+
+  public ApiFuture<Void> run() {
+    Preconditions.checkState(
+        documentsPending, "The recursive delete operation has already been completed");
```
schmidt-sebastian, line 131: This seems like it is not the correct check. We should only allow this call once. It should not be called again if it is already running, whether or not it is completed. You can either fix this check or remove it.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
schmidt-sebastian: We probably need to add a `verifyNotClosed` that calls `verifyNotClosedLocked`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+   * query to fetch descendants. By starting the query while there are pending operations, Firestore
+   * can improve BulkWriter throughput. This helps prevent BulkWriter from idling while Firestore
+   * fetches the next query.
+   */
+  public static final int MIN_PENDING_OPS = 1000;
+
+  private final FirestoreRpcContext<?> firestoreRpcContext;
+  private final BulkWriter writer;
+  @Nullable private final CollectionReference collectionReference;
+  @Nullable private final DocumentReference documentReference;
```
schmidt-sebastian, line 69: Can this be simplified if we just pass "parentPath"?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+        new ApiStreamObserver<DocumentSnapshot>() {
+          public void onNext(DocumentSnapshot snapshot) {
+            streamedDocsCount[0]++;
+            lastDocumentSnap = snapshot;
+            deleteReference(snapshot.getReference());
+          }
+
+          public void onError(Throwable throwable) {
+            lastError =
+                new FirestoreException("Failed to fetch children documents.", Status.UNAVAILABLE);
```
schmidt-sebastian, line 164: This swallows the underlying exception.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+                .setAllDescendants(true)
+                .setKindless(true)
+                .setRequireConsistency(false)
+                .build());
+
+    // Query for names only to fetch empty snapshots.
+    query = query.select(FieldPath.documentId()).limit(maxPendingOps);
+
+    return query;
+  }
```
schmidt-sebastian, line 240: We should find a way to combine all three methods (by passing ResourcePath and collectionId at each callsite)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+  private void onQueryEnd() {
+    documentsPending = false;
+
+    // Used to aggregate flushFuture and deleteFuture to use with ApiFutures.allAsList(), in order
+    // to ensure that the delete catchingAsync() callback is run before the flushFuture callback.
+    List<ApiFuture<Void>> pendingFutures = new ArrayList<>();
+
+    // Delete the provided document reference if one was provided.
+    if (documentReference != null) {
+      ApiFuture<Void> voidDeleteFuture =
```
schmidt-sebastian, line 256: Can we call `deleteReference`? (Not sure)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+
+  /** Deletes the provided reference and starts the next stream if conditions are met. */
+  private void deleteReference(final DocumentReference reference) {
+    pendingOperationsCount++;
+    ApiFutures.transformAsync(
+        catchingDelete(reference),
+        new ApiAsyncFunction<WriteResult, Object>() {
+
+          public ApiFuture<Object> apply(WriteResult result) {
+            pendingOperationsCount--;
```
schmidt-sebastian, line 321: If BulkWriter sends two requests in parallel, this code might be invoked from two different Gax threads. I think this needs synchronization (same with the error handler).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,353 @@
...
+
+    ApiFuture<Void> flushFuture = writer.flush();
+    pendingFutures.add(flushFuture);
+
+    // Completes the future returned by run() and sets the exception if an error occurred.
+    ApiFutures.transformAsync(
+        ApiFutures.allAsList(pendingFutures),
+        new ApiAsyncFunction<List<Void>, Void>() {
+          public ApiFuture<Void> apply(List<Void> unused) {
+            if (lastError == null) {
```
schmidt-sebastian, line 276: Is this guaranteed to run on the same thread that sets `lastError`?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/RecursiveDeleteTest.java
```diff
@@ -0,0 +1,678 @@
...
+    FirestoreException mockException =
+        new FirestoreException("runQuery() failed in test", Status.UNAVAILABLE);
+    // Note that these mock responses differ from the Node implementation since Node retries
+    // queries that fail without streaming any documents in the SDK, whereas Java handles these
+    // retries at the GAX level. The Java SDK does not retry unless a document was streamed in the
+    // query.
+    final List<Answer<RunQueryResponse>> runQueryResponses =
+        new ArrayList<>(
+            Arrays.asList(
+                queryResponse(mockException, fullDocumentPath("a"), fullDocumentPath("b")),
```
schmidt-sebastian, line 564: Wouldn't the exception be the last response rather than the first?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/RecursiveDeleteTest.java
```diff
@@ -0,0 +1,678 @@
...
+        .when(firestoreMock)
+        .bulkWriter();
+
+    firestoreMock.recursiveDelete(firestoreMock.collection("foo")).get();
+    firestoreMock.recursiveDelete(firestoreMock.collection("boo")).get();
+    firestoreMock.recursiveDelete(firestoreMock.collection("moo")).get();
+
+    // Only the first recursiveDelete() call should have called the constructor. Subsequent calls
+    // should have used the same bulkWriter.
+    assertEquals(1, callCount[0]);
```
schmidt-sebastian, line 664: Very clever.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1686,4 +1688,110 @@ public void testBuildingBundleWithLimitToLastQuery() throws Exception {
...
+  public void testRecursiveDeleteDoesNotAffectOtherCollections() throws Exception {
+    setupRecursiveDeleteTest();
+
+    // Add another nested collection that shouldn't be deleted.
+    CollectionReference collectionB = firestore.collection("doggos");
+    collectionB.document("doggo").set(map("name", "goodboi"));
+
+    firestore.recursiveDelete(collectionB).get();
+    assertEquals(6, countCollectionChildren(randomColl));
+    assertEquals(0, countCollectionChildren(collectionB));
```
schmidt-sebastian, line 1777: Should there not still be one document?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/RecursiveDeleteTest.java
```diff
@@ -0,0 +1,678 @@
...
+    bulkWriter = firestoreMock.bulkWriter();
+    bulkWriter.close();
+    try {
+      firestoreMock.recursiveDelete(firestoreMock.collection("foo"), bulkWriter);
+      fail("Operation should have failed in test");
+    } catch (Exception e) {
+      assertTrue(e.getMessage().contains("BulkWriter has already been closed"));
+    }
+  }
+}
```
schmidt-sebastian, line 678: I know it will be pretty tough to test but we do not have a test that tests that we send the second request when we pass the MIN_PENDING_OPS threshold.

schmidt-sebastian (commented):
Looks great!

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
thebrianchen: Wrapped the old exception. The status should be `UNAVAILABLE` regardless of the error, as it is on Node.

thebrianchen (commented):
thanks for the review!

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -102,6 +117,44 @@ public BulkWriter bulkWriter(BulkWriterOptions options) {
...
+
+  /**
+   * This overload is only used to test the query resumption with startAfter() once the
+   * RecursiveDelete instance has MAX_PENDING_OPS pending.
+   */
+  @Nonnull
+  @VisibleForTesting
+  ApiFuture<Void> recursiveDelete(
+      CollectionReference reference, BulkWriter bulkWriter, int maxLimit, int minLimit) {
+    RecursiveDelete deleter = new RecursiveDelete(this, bulkWriter, reference, maxLimit, minLimit);
```
schmidt-sebastian, line 154: This code flow would be a bit cleaner if you called this method from all overloads. This will ensue that all calls go through the same callsite. If you take `parentPath` and `collectionId` as input here then you will also remove some duplication in RecursiveDelete (as you only need one constructor).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,356 @@
...
+  private boolean invoked = false;
+
+  /** Query limit to use when fetching all descendants. */
+  private int maxPendingOps = MAX_PENDING_OPS;
+
+  /**
+   * The number of pending BulkWriter operations at which RecursiveDelete starts the next limit
+   * query to fetch descendants.
+   */
+  private int minPendingOps = MIN_PENDING_OPS;
```
schmidt-sebastian, line 110: This can be final now.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,356 @@
...
+
+  /** Whether a query stream is currently in progress. Only one stream can be run at a time. */
+  @GuardedBy("lock")
+  private boolean streamInProgress = false;
+
+  /** Whether run() has been called. */
+  private boolean invoked = false;
+
+  /** Query limit to use when fetching all descendants. */
+  private int maxPendingOps = MAX_PENDING_OPS;
```
schmidt-sebastian, line 104: This can be final now.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,356 @@
...
+
+  /** A deferred promise that resolves when the recursive delete operation is completed. */
+  private final SettableApiFuture<Void> completionFuture = SettableApiFuture.create();
+
+  /** Whether a query stream is currently in progress. Only one stream can be run at a time. */
+  @GuardedBy("lock")
+  private boolean streamInProgress = false;
+
+  /** Whether run() has been called. */
+  private boolean invoked = false;
```
schmidt-sebastian, line 101: started?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,356 @@
...
+   * query to fetch descendants. By starting the query while there are pending operations, Firestore
+   * can improve BulkWriter throughput. This helps prevent BulkWriter from idling while Firestore
+   * fetches the next query.
+   */
+  public static final int MIN_PENDING_OPS = 1000;
+
+  private final FirestoreRpcContext<?> firestoreRpcContext;
+  private final BulkWriter writer;
+
+  @Nullable private final DocumentReference documentReference;
```
schmidt-sebastian, line 69: It might be cleaner to replace this with a documentId string. Then you would have parentPath, collectionId and documentId below. What do you think?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,356 @@
...
+            synchronized (lock) {
+              String message = "Failed to fetch children documents. " + throwable.getMessage();
+              lastError = new FirestoreException(message, Status.UNAVAILABLE);
+              onQueryEnd();
+            }
+          }
+
+          public void onCompleted() {
+            synchronized (lock) {
+              streamInProgress = false;
```
schmidt-sebastian, line 199: We should probably reduce the scope of the lock to just this line (also for the onError callback, which only needs to set lastError under lock). This improves readability as it makes the intention more clear and also answer the question why `onNext` does not use locks.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
schmidt-sebastian: Got it. I am glad you made me look at this again though - I think you should wait for "doggo" to be written before invoking the delete.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/RecursiveDeleteTest.java
```diff
@@ -0,0 +1,733 @@
...
+            streamObserverCapture.capture(),
+            Matchers.<ServerStreamingCallable>any());
+
+    doAnswer(
+            new Answer<ApiFuture<BatchWriteResponse>>() {
+              public ApiFuture<BatchWriteResponse> answer(InvocationOnMock mock) throws Exception {
+                if (numDeletesBuffered[0] < cutoff) {
+                  numDeletesBuffered[0] += batchWriteResponse.size();
+                  return ApiFutures.transformAsync(
+                      bufferFuture,
```
schmidt-sebastian, line 393: I don't quite understand why we need to block on `bufferFuture` here. I thought the idea was that we should finish the first 4000 deletes and then ask for the next batch. The current test seems to indicate that we do not wait for the deletes to finish executing.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/RecursiveDeleteTest.java
```diff
@@ -0,0 +1,733 @@
...
+                } else {
+                  // Once there are `cutoff` pending deletes, completing the future allows enough
+                  // responses to be returned such that the number of pending deletes should be less
+                  // than `minPendingOps`. This allows us to test that the second query is made.
+                  bufferFuture.set(null);
+                  return ApiFutures.transformAsync(
+                      secondQueryFuture,
+                      new ApiAsyncFunction<Void, BatchWriteResponse>() {
+                        public ApiFuture<BatchWriteResponse> apply(Void unused) throws Exception {
+                          System.out.println(
```
schmidt-sebastian, line 409: Please remove

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
schmidt-sebastian: You can make it return an ApiFuture.

```
private ApiFuture<Void> deleteReference(final DocumentReference reference) {
    synchronized (lock) {
      pendingOperationsCount++;
    }
    
    return ApiFutures.transformAsync(
          catchingDelete(reference),
          new ApiAsyncFunction<WriteResult, Void>() {

            public ApiFuture<Void> apply(WriteResult result) {
              synchronized (lock) {
                pendingOperationsCount--;
                // We wait until the previous stream has ended in order to ensure the
                // startAfter document is correct. Starting the next stream while
                // there are pending operations allows Firestore to maximize
                // BulkWriter throughput.
                if (documentsPending
                    && !streamInProgress
                    && pendingOperationsCount < minPendingOps) {
                  streamDescendants();
                }
                return ApiFutures.immediateFuture(null);
              }
            }
          },
          MoreExecutors.directExecutor());
    }
  }
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -65,27 +65,41 @@
...
   /** Whether there are still documents to delete that still need to be fetched. */
+  @GuardedBy("lock")
   private boolean documentsPending = true;
 
   /** A deferred promise that resolves when the recursive delete operation is completed. */
   private final SettableApiFuture<Void> completionFuture = SettableApiFuture.create();
 
   /** Whether a query stream is currently in progress. Only one stream can be run at a time. */
+  @GuardedBy("lock")
   private boolean streamInProgress = false;
```
schmidt-sebastian, line 98: You might be able to combine `streamInProgress` and `documentsPending` with some `!` magic.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -113,42 +127,55 @@
...
-            || (documentReference != null && collectionReference == null),
-        "Either documentReference or collectionReference should be set");
-
-    Query query;
-    if (collectionReference != null) {
-      query = getAllDescendantsQuery(collectionReference);
-    } else {
-      query = getAllDescendantsQuery(documentReference);
-    }
+
```
schmidt-sebastian, line 177: Nit: Empty line.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -160,48 +187,59 @@ public void onNext(DocumentSnapshot snapshot) {
           }
 
           public void onError(Throwable throwable) {
-            lastError =
-                new FirestoreException("Failed to fetch children documents.", Status.UNAVAILABLE);
-            onQueryEnd();
+            synchronized (lock) {
+              String message = "Failed to fetch children documents. " + throwable.getMessage();
+              lastError = new FirestoreException(message, Status.UNAVAILABLE);
```
schmidt-sebastian, line 192: Java exception "bubbling" usually involves passing the original exception as the cause to the wrapping exception.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
thebrianchen: I think trying to rehydrate the DocumentReference from the documentId is more bug-prone, since we'd be setting it to an empty string for collections, and the document path for documents.

I renamed `documentReference` to `rootDocumentReference` and added documentation. This way, we don't have to add extra code to rehydrate, and it's pretty clear to see the purpose of the variable at a glance.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
thebrianchen: Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
thebrianchen: Ah, I see what you mean now, thanks. Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
```diff
@@ -0,0 +1,331 @@
...
+
+  private Query getAllDescendantsQuery() {
+    ResourcePath parentPath;
+    String collectionId;
+    if (path.isDocument()) {
+      // The parent is the closest ancestor document to the location we're deleting. Since we are
+      // deleting a document, the parent is the path of that document.
+      parentPath = path;
+      collectionId = path.getParent().getId();
+      Preconditions.checkState(collectionId != null, "Parent of a document should not be null.");
```
schmidt-sebastian, line 193: I would drop this check - every document has a collection ID. If you do keep it, the check should be on getParent(), which is nullable and not on getId(). 

BTW, I really like this cleanup.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/RecursiveDeleteTest.java
schmidt-sebastian: Thanks for explaining this. I am pretty sure I understand this now. If you can add a small comment in the test I might even still understand tomorrow (optional).

schmidt-sebastian (approved):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/RecursiveDelete.java
thebrianchen: Changed check to parent.

I definitely like this version more, thanks for pushing through on your vision :) Having all the collectionId/parentPath logic inside getAllDescendantsQuery() makes so more sense, and makes the rest of the code much cleaner!

google-cloud-firestore/src/test/java/com/google/cloud/firestore/RecursiveDeleteTest.java
thebrianchen: Added a comment.

PR #631, opened 2021-05-14
fix: add flush check in BulkWriter.sendBatchLocked() callback

thebrianchen:
Found this bug while adding recursive delete tests. I tried to think of a way to add a unit test to check for this, but I couldn't think of an easy test since `scheduleCurrentBatchLocked()` automatically returns if there are no mutations.

However, not having this check was causing the `startAfter()` recursive delete test to flake, so that's some indirect coverage 😨

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
@@ -989,7 +989,9 @@ public final class BulkWriter implements AutoCloseable {
                 @Override
                 public void run() {
                   synchronized (lock) {
-                    scheduleCurrentBatchLocked(flush);
+                    if (flush) {
+                      scheduleCurrentBatchLocked(/* flush= */ true);
+                    }
                   }
                 }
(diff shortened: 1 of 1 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -989,7 +989,9 @@ private void sendBatchLocked(final BulkCommitBatch batch, final boolean flush) {
                 @Override
                 public void run() {
                   synchronized (lock) {
-                    scheduleCurrentBatchLocked(flush);
+                    if (flush) {
```
schmidt-sebastian, line 992: The if-check can happen outside the synchronized block.

schmidt-sebastian (approved):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: done.

PR #653, opened 2021-05-27
fix: return results from getPartitions() in order

schmidt-sebastian:
The Partition RPC does not guarantee that results are returned in order, which means we have to sort the partitions before we can use them as cursors.

Port of https://github.com/googleapis/nodejs-firestore/pull/1521

We don't have unit tests for PartitionQueries as we never figured out how to mock paged respones.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
@@ -65,8 +65,8 @@ public class CollectionGroup extends Query {
    * parallel. The returned partition cursors are split points that can be used as starting/end
    * points for the query results.
    *
-   * @deprecated Please use {@link #getPartitions(long)} instead. All cursors will be loaded before any value will be provided to {@code observer}.
-   *     the server. Use {@link #getPartitions(long)} to get the same behavior.
+   * @deprecated Please use {@link #getPartitions(long)} instead. All cursors will be loaded before
+   *     any value will be provided to {@code observer}.
    * @param desiredPartitionCount The desired maximum number of partition points. The number must be
    *     strictly positive. The actual number of partitions returned may be fewer.
    * @param observer a stream observer that receives the result of the Partition request.
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
@@ -65,7 +65,7 @@ public class CollectionGroup extends Query {
    * parallel. The returned partition cursors are split points that can be used as starting/end
    * points for the query results.
    *
-   * @deprecated This method is deprecated as the SDK has to post-process all results streamed from
+   * @deprecated Please use {@link #getPartitions(long)} instead. All cursors will be loaded before any value will be provided to {@code observer}.
    *     the server. Use {@link #getPartitions(long)} to get the same behavior.
    * @param desiredPartitionCount The desired maximum number of partition points. The number must be
    *     strictly positive. The actual number of partitions returned may be fewer.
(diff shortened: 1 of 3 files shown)


kolea2 (commented):
Would this example of mocking paged responses help? Maybe we could add something similar here? https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientTest.java#L647-L663

schmidt-sebastian:
@kolea2 It might - I will see if I have some cycles to flush this out. I think we should not block this bug fix on it though, as the code does get executed by the existing integration tests.

BenWhitehead:
Mocking of PartitionQuery paged responses is present in my beam work https://github.com/BenWhitehead/beam/blob/66c203fced82b938c2251d21fa45e04a6b6e42d9/sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/firestore/FirestoreV1FnPartitionQueryTest.java#L48

schmidt-sebastian:
Thanks for the pointers. I added unit tests.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
```diff
@@ -63,6 +65,8 @@
    * parallel. The returned partition cursors are split points that can be used as starting/end
    * points for the query results.
    *
+   * @deprecated This method is deprecated as the SDK has to post-process all results streamed from
```
crwilcox, line 68: maybe non-ideal, but could we serve the results of `getPartitions(long)` as a stream here, ensuring ordering? It won't actually be streaming of course.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/PartitionQuery.java
```diff
@@ -0,0 +1,204 @@
...
+            Matchers.<UnaryCallable<PartitionQueryRequest, PartitionQueryPagedResponse>>any());
+
+    firestoreMock.collectionGroup("collectionId").getPartitions(desiredPartitionsCount).get();
+
+    PartitionQueryRequest actualRequest = requestCaptor.getValue();
+    assertEquals(actualRequest, expectedRequest);
+  }
+
+  @Test
+  public void doesNotIssueRpcIfOnlyASinglePartitionIsRequested() throws Exception {
```
crwilcox, line 121: This seems like an interesting optimization. And also like an unlikely case? Is it worth special handling in libs or should we just let the backend respond with no partitions?

crwilcox (approved):

google-cloud-firestore/src/test/java/com/google/cloud/firestore/PartitionQuery.java
schmidt-sebastian: The backend returns an error when 0 partitions are requested. We should catch this ourselves since we request one less than what the user requested.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
schmidt-sebastian: We could and I would like more feedback on this decision. I feel it sends the wrong message though to keep the streaming API when it just buffers.

cc @BenWhitehead

crwilcox: My concern is that even deprecated, the result is incorrect? I do think streaming sends the wrong message.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/PartitionQuery.java
crwilcox: I had already ensure we always returned a minimum of 1 QueryPartition. I just added the short circuit in golang though, since it is the case we don't really need to request anything.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
schmidt-sebastian: The result is fetched, then sorted, then released to the user so the behavior should be correct.

BenWhitehead: Discouraging folks I think is good here since the behavioral semantics have changed from what was initially expected. 

I might phrase it differently though
```suggestion
   * @deprecated Please use {@link #getPartitions(long)} instead. All cursors will be loaded before any value will be provided to {@code observer}.
```

We also want to add the `@Deprecated` annotation to the method itself.

PR #670, opened 2021-06-18
test(fix): remove assertion on error message for reads older than 60 seconds

kolea2:
It seems like the error message for this has been updated.

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
@@ -1537,7 +1537,6 @@ public class ITSystemTest {
       final StatusRuntimeException invalidArgument = (StatusRuntimeException) rootCause;
       final Status status = invalidArgument.getStatus();
       assertThat(status.getCode()).isEqualTo(Code.FAILED_PRECONDITION);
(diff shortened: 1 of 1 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1537,7 +1537,7 @@ public Void updateCallback(Transaction transaction) throws Exception {
       final StatusRuntimeException invalidArgument = (StatusRuntimeException) rootCause;
       final Status status = invalidArgument.getStatus();
       assertThat(status.getCode()).isEqualTo(Code.FAILED_PRECONDITION);
-      assertThat(status.getDescription()).contains("minimum");
+      assertThat(status.getDescription()).isEqualTo("The requested snapshot version is too old.");
```
kolea2, line 1540: Question to reviewer: I noticed we've updated this assertion before - is it worth keeping (is the previous assertion enough?)

wu-hui: I think we can just delete this line, checking for exact messages does not bring much value and is easy to break.

kolea2: Thanks @wu-hui, I've removed the assertion.

wu-hui (approved):

PR #671, opened 2021-06-18
fix: handle thrown exceptions in runAsyncTransaction callback

### LINKED ISSUE #667: firestore-admin: runAsyncTransaction() throwing exception never stops retrying
OS: Linux
openjdk version 11.0.11
firebase-admin:7.3.0
google-cloud-firestore:2.5.0

A transaction with the local emulator that always throws an exception is constantly retried and never times out:

```java
var options = FirestoreOptions
  .getDefaultInstance()
  .toBuilder()
  .setProjectId(Config.GOOGLE_CLOUD_PROJECT)
  .setEmulatorHost("localhost:9000")
  .setCredentials(new FirestoreOptions.EmulatorCredentials());
var firestore = options.build().getService();

firestore
  .runAsyncTransaction(
    tx -> {
      log.info("Hello world");
      throw new RuntimeException("Fail");
      // return ApiFutures.immediateFuture("OK");
    }
  )
  .get();
```

If I replace the `throw` statement with `return`, the invocation finishes successfully. 

I also tried adding the following, and it didn't work:

```java
options.setRetrySettings(
  RetrySettings
    .newBuilder()
    .setMaxAttempts(1)
    .setTotalTimeout(Duration.ofSeconds(2))
    .setInitialRpcTimeout(Duration.ofSeconds(2))
    .setMaxRpcTimeout(Duration.ofSeconds(2))
    .build()
)
```

I expected the following:

* Transaction retries to expire after 60 seconds
* With the second example, the transaction to stop retrying after one failed attempt or 2 seconds passed.

Did I miss something?

thebrianchen:
Fixes #667.

The callback `onFailure` is only triggered if the ApiFuture contains the exception. When a runtime exception is thrown neither the `onFailure` nor `onSuccess` handler is called, leading the ApiFuture to hang and never complete.

The fix adds a `catchingAsync()` around `updateCallback` to guard against thrown exceptions. Added a system test for redundancy, but I can remove it. Both tests fail (never complete) when run against master.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
@@ -145,33 +145,34 @@ class TransactionRunner<T> {
    * Invokes the user callback on the user callback executor and returns the user-provided result.
    */
   private SettableApiFuture<T> invokeUserCallback() {
-    final SettableApiFuture<T> callbackResult = SettableApiFuture.create();
+    final SettableApiFuture<T> returnedResult = SettableApiFuture.create();
 
     userCallbackExecutor.execute(
         new Runnable() {
           @Override
           public void run() {
             try {
+              ApiFuture<T> userCallbackResult = userCallback.updateCallback(transaction);
               ApiFutures.addCallback(
-                  userCallback.updateCallback(transaction),
+                  userCallbackResult,
                   new ApiFutureCallback<T>() {
                     @Override
                     public void onFailure(Throwable t) {
-                      callbackResult.setException(t);
+                      returnedResult.setException(t);
                     }
 
                     @Override
                     public void onSuccess(T result) {
-                      callbackResult.set(result);
+                      returnedResult.set(result);
                     }
                   },
                   firestoreExecutor);
             } catch (Exception e) {
-              callbackResult.setException(e);
+              returnedResult.setException(e);
             }
           }
         });
-    return callbackResult;
+    return returnedResult;
   }
 
   /** A callback that invokes the BeginTransaction callback. */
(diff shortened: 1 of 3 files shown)


schmidt-sebastian (commented):
Did you look at https://github.com/googleapis/java-firestore/blob/f2c76a15061984327547715401c93e17f9838113/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java#L465? It might be cleaner to follow the same pattern here.

thebrianchen:
> ? It might be cleaner to follow the same pattern here.

I saw that, but it's only clean for synchronous functions. For async callbacks that return an ApiFuture, I think the only way is to add a callback.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/TransactionRunner.java
```diff
@@ -146,24 +146,29 @@ public void run() {
...
-            ApiFutures.addCallback(
-                userCallback.updateCallback(transaction),
-                new ApiFutureCallback<T>() {
-                  @Override
-                  public void onFailure(Throwable t) {
-                    callbackResult.setException(t);
-                  }
+            try {
+              ApiFutures.addCallback(
+                  userCallback.updateCallback(transaction),
```
schmidt-sebastian, line 156: I think this code would be easier to understand if you did something like this:

```
try {
const result = userCallback.updateCallback(transaction)
ApiFutures.addCallback(...)
} catch (Exception e) {
   callbackResult.setException(e);
}
```

It's very hard to tell what the try/catch block addresses. The other alternative is to wrap the callback in a try/catch before passing it here like in TransactionAsyncFunction.

thebrianchen: Done, moved it out into a separate variable.

schmidt-sebastian (approved):

PR #688, opened 2021-07-02
fix: lower batch size on BulkWriter retry

thebrianchen:


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -82,6 +83,9 @@
   /** The maximum number of writes that can be in a single batch. */
   public static final int MAX_BATCH_SIZE = 20;
 
+  /** The maximum number of writes can be can in a single batch that is being retried. */
```
schmidt-sebastian, line 86: This is Sebastian-level grammar. Please upgrade to Brian-level grammar.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -1050,7 +1057,13 @@ void setMaxPendingOpCount(int newMax) {
   private void sendOperationLocked(
       ApiFunction<BulkCommitBatch, ApiFuture<WriteResult>> enqueueOperationOnBatchCallback,
       final BulkWriterOperation op) {
-    if (bulkCommitBatch.has(op.getDocumentReference())) {
+    // A backoff duration greater than 0 implies that this batch is a retry.
+    // Retried writes are sent with a batch size of 10 in order to guarantee
+    // that the batch is under the 10MiB limit.
+    boolean retryInSmallerBatch =
```
schmidt-sebastian, line 1063: This boolean does more than just check whether a smaller batch should be used. It also checks the size of the current batch. I think we need to move some logic or change the name.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -1062,7 +1075,11 @@ private void sendOperationLocked(
     bulkCommitBatch.enqueueOperation(op);
     enqueueOperationOnBatchCallback.apply(bulkCommitBatch);
 
-    if (bulkCommitBatch.getMutationsSize() == maxBatchSize) {
+    if (op.getBackoffDuration() > 0) {
```
schmidt-sebastian, line 1078: I am a bit confused. Why do we need this and `retryInSmallerBatch`? Is it possible to remove one of these checks?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
```diff
@@ -1062,7 +1075,11 @@ private void sendOperationLocked(
...
+      bulkCommitBatch.setMaxBatchSize(RETRY_MAX_BATCH_SIZE);
+    }
+
+    if (bulkCommitBatch.getMutationsSize() == bulkCommitBatch.getMaxBatchSize()) {
```
schmidt-sebastian, line 1082: ... especially since we are reading the value right back here.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: The first check involving `retryInSmallerBatch` ensures that we are adding the retry to an appropriate batch. For example, if the batch already had 15 writes enqueued, and `sendOperationLocked()` is called with a retry, BulkWriter needs to first send the existing batch to create a new batch of size 10.

Consolidated the checks into a single if statement at the start of the function.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkWriter.java
thebrianchen: Removed.

schmidt-sebastian (approved):

PR #726, opened 2021-08-10
chore: hello Java8

schmidt-sebastian:
This PR is the result of running Java8 migrations via IntelliJ (with some occasional fine-tuning).

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
@@ -16,8 +16,6 @@
 
 package com.google.cloud.firestore;
 
-import com.google.api.core.ApiAsyncFunction;
-import com.google.api.core.ApiFunction;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutures;
 import com.google.api.gax.rpc.ApiException;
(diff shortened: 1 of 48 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
```diff
@@ -127,20 +122,17 @@ void setMaxBatchSize(int size) {
...
+          com.google.rpc.Status.Builder status =
+              com.google.rpc.Status.newBuilder()
+                  .setCode(exception.getStatusCode().getCode().ordinal())
+                  .setMessage(exception.getMessage());
+          BatchWriteResponse.Builder response1 = BatchWriteResponse.newBuilder();
+          for (int i = 0; i < pendingOperations.size(); ++i) {
+            response1.addWriteResults(com.google.firestore.v1.WriteResult.getDefaultInstance());
+            response1.addStatus(status);
           }
+          return response1.build();
```
kolea2, line 135: nit - would rename this to something more descriptive, ie `batchWriteResponse`

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
```diff
@@ -305,7 +305,7 @@ static Object convertToPlainJavaTypes(Object object) {
       try {
         result =
             (rawType == Map.class)
-                ? new HashMap<String, Object>()
+                ? new HashMap<>()
```
kolea2, line 308: I know it's not needed, but do you think we could gain any readability here from being explicit?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/ResourcePath.java
```diff
@@ -185,11 +185,6 @@ ResourcePath createPathWithSegments(ImmutableList<String> segments) {
   }
 
   public static Comparator<ResourcePath> comparator() {
-    return new Comparator<ResourcePath>() {
-      @Override
-      public int compare(ResourcePath left, ResourcePath right) {
-        return left.compareTo(right);
-      }
-    };
+    return Comparator.naturalOrder();
```
kolea2, line 188: just want to confirm this is tested so there's no surprises

kolea2 (approved):
LGTM with a few questions/nits

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
kolea2: nvm, the code above was hidden. This is fine 👍

google-cloud-firestore/src/main/java/com/google/cloud/firestore/ResourcePath.java
schmidt-sebastian: The code is equivalent as it ends up calling `compareTo`. The tests for this are in https://github.com/googleapis/java-firestore/blob/c5e43aacb1c8977b475b9f3eac9be6ab79c31975/google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java

schmidt-sebastian (commented):
Thanks for the review!

google-cloud-firestore/src/main/java/com/google/cloud/firestore/ResourcePath.java
kolea2: sgtm!

Issue #822, opened 2021-11-29
Snapshot listener gets a regular callback with 'Received unexpected EOS on empty DATA frame from server' exception


menulis:
We are using Firestore snapshot listeners on documents and queries to receive callbacks whenever documents are updated.
The [documentation](https://cloud.google.com/firestore/docs/query-data/listen#java_6) states that "after an error, the listener will not receive any more events, and there is no need to detach your listener“. However, once a snapshot listener is created, we regularly  (around every hour at least for documents that do not get any updates) receive an error callback, i.e. `EventListener::onEvent` with non-null second `error` argument that is of `FirestoreException` type that looks like a gRPC idle timeout or similar:

```
io.grpc.StatusRuntimeException: INTERNAL: Received unexpected EOS on empty DATA frame from server
	at io.grpc.Status.asRuntimeException(Status.java:535)
	... 10 common frames omitted
Wrapped by: com.google.api.gax.rpc.InternalException: io.grpc.StatusRuntimeException: INTERNAL: Received unexpected EOS on empty DATA frame from server
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:67)
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:72)
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:60)
	at com.google.api.gax.grpc.ExceptionResponseObserver.onErrorImpl(ExceptionResponseObserver.java:82)
	at com.google.api.gax.rpc.StateCheckingResponseObserver.onError(StateCheckingResponseObserver.java:86)
	at com.google.api.gax.grpc.GrpcDirectStreamController$ResponseObserverAdapter.onClose(GrpcDirectStreamController.java:149)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:557)
	at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:69)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:738)
	at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:717)
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
[...]

PR #834, opened 2021-12-20
fix: close the Watch stream when we receive an error

schmidt-sebastian:
The Watch code called "onCompleted" to close the stream - but that did not actually do anything. This PR migrates our streaming GAPIC APIs to ClientStream (which is not deprecated) and then calls "close()" instead.

Fixes: https://github.com/googleapis/java-firestore/issues/822

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreRpcContext.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreRpcContext.java
@@ -20,8 +20,10 @@ import com.google.api.core.ApiClock;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.InternalApi;
 import com.google.api.core.InternalExtensionOnly;
-import com.google.api.gax.rpc.ApiStreamObserver;
+import com.google.api.gax.rpc.BidiStreamObserver;
 import com.google.api.gax.rpc.BidiStreamingCallable;
+import com.google.api.gax.rpc.ClientStream;
+import com.google.api.gax.rpc.ResponseObserver;
 import com.google.api.gax.rpc.ServerStreamingCallable;
 import com.google.api.gax.rpc.UnaryCallable;
 import com.google.cloud.firestore.spi.v1.FirestoreRpc;
@@ -48,10 +50,10 @@ interface FirestoreRpcContext<FS extends Firestore> {
 
   <RequestT, ResponseT> void streamRequest(
       RequestT requestT,
-      ApiStreamObserver<ResponseT> responseObserverT,
+      ResponseObserver<ResponseT> responseObserverT,
       ServerStreamingCallable<RequestT, ResponseT> callable);
 
-  <RequestT, ResponseT> ApiStreamObserver<RequestT> streamRequest(
-      ApiStreamObserver<ResponseT> responseObserverT,
+  <RequestT, ResponseT> ClientStream<RequestT> streamRequest(
+      BidiStreamObserver<RequestT, ResponseT> responseObserverT,
       BidiStreamingCallable<RequestT, ResponseT> callable);
 }
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
@@ -21,7 +21,9 @@ import com.google.api.gax.grpc.GrpcStatusCode;
 import com.google.api.gax.retrying.ExponentialRetryAlgorithm;
 import com.google.api.gax.retrying.TimedAttemptSettings;
 import com.google.api.gax.rpc.ApiException;
-import com.google.api.gax.rpc.ApiStreamObserver;
+import com.google.api.gax.rpc.BidiStreamObserver;
+import com.google.api.gax.rpc.ClientStream;
+import com.google.api.gax.rpc.StreamController;
 import com.google.cloud.Timestamp;
 import com.google.cloud.firestore.DocumentChange.Type;
 import com.google.common.base.Preconditions;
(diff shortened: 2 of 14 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreRpcContext.java
```diff
@@ -48,10 +50,10 @@
...
   <RequestT, ResponseT> void streamRequest(
       RequestT requestT,
-      ApiStreamObserver<ResponseT> responseObserverT,
+      ResponseObserver<ResponseT> responseObserverT,
       ServerStreamingCallable<RequestT, ResponseT> callable);
 
-  <RequestT, ResponseT> ApiStreamObserver<RequestT> streamRequest(
-      ApiStreamObserver<ResponseT> responseObserverT,
+  <RequestT, ResponseT> ClientStream<RequestT> streamRequest(
+      BidiStreamObserver<RequestT, ResponseT> responseObserverT,
```
schmidt-sebastian, line 57: These are the main changes, the rest is plumbing and cleanup.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java
```diff
@@ -289,7 +297,7 @@ ListenerRegistration runWatch(
           .execute(
               () -> {
                 synchronized (Watch.this) {
-                  stream.onCompleted();
+                  stream.closeSend();
```
schmidt-sebastian, line 300: This now closes the stream instead of invoking a no-op method.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
```diff
@@ -957,20 +957,19 @@ public void retriesAfterRetryableError() throws Exception {
...
+                return queryResponse(
+                        FirestoreException.forServerRejection(
+                            Status.DEADLINE_EXCEEDED, "Simulated test failure"),
+                        DOCUMENT_NAME + "1",
+                        DOCUMENT_NAME + "2")
+                    .answer(invocation);
+              } else {
+                return queryResponse(DOCUMENT_NAME + "3").answer(invocation);
+              }
+            })
```
schmidt-sebastian, line 972: FYI: Here and below - this is only a formatting change as I removed the unused cast.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/RecursiveDeleteTest.java
```diff
@@ -355,48 +356,46 @@ public void createsSecondQueryWithCorrectStartAfter() throws Exception {
...
-                    // made.
-                    bufferFuture.set(null);
-                    return ApiFutures.transformAsync(
-                        secondQueryFuture,
-                        (ApiAsyncFunction<Void, BatchWriteResponse>)
-                            unused -> mergeResponses(batchWriteResponse.toArray(new ApiFuture[0])),
-                        MoreExecutors.directExecutor());
-                  }
-                })
+            mock -> {
```
schmidt-sebastian, line 374: Only a formatting change.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
```diff
@@ -1119,11 +1118,10 @@ public void failsWritesAfterAllRetryAttemptsFail() throws Exception {
         };
 
     doAnswer(
-            (Answer<ApiFuture<GeneratedMessageV3>>)
-                mock -> {
-                  retryAttempts[0]++;
-                  return RETRYABLE_FAILED_FUTURE;
-                })
+            mock -> {
```
kolea2, line 1121: will this test the new changes?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/ConformanceTest.java
```diff
@@ -641,11 +641,10 @@ public void runTest() throws Throwable {
       final SettableApiFuture<Void> testCaseFinished = SettableApiFuture.create();
 
       doAnswer(
-              (Answer<ApiStreamObserver<ListenRequest>>)
-                  invocationOnMock -> {
-                    testCaseStarted.set(null);
-                    return noOpRequestObserver;
-                  })
+              invocationOnMock -> {
```
kolea2, line 644: same q here

kolea2 (commented):
changes look fine. Just a few questions around testing

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
schmidt-sebastian: No, it's hard to test these changes. We could mock the backend stream, but then we are essentially only testing the behavior of the mock. If you know of a pre-existing implementation/fake of a GRPC stream that we can use to test this behavior, then we can add a test. A homegrown implementation that validates that our code follows our ow assumptions will not provide us with much meaningful test coverage.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/ConformanceTest.java
schmidt-sebastian: These are just Lambda cleanups.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/BulkWriterTest.java
kolea2: Would something like this be helpful? https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/gaxx/testing/FakeStreamingApi.java
Examples in tests:
https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/FilterMarkerRowsCallableTest.java#L38

(It's using ApiStreamObserver, but I imagine could be refactored to this use case)

schmidt-sebastian: Added test queryWatchShutsDownStreamOnPermissionDenied() that re-uses some of the existing functionality.

kolea2 (approved):

PR #870, opened 2022-02-11
tests: add coverage for a successful read-only transaction with readTime

cherylEnkidu:
Add a test to cover the a successful transaction with readTime set before the last time changing the document.

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
@@ -1491,6 +1491,37 @@ public class ITSystemTest {
   }
 
   @Test
+  public void readOnlyTransaction_successfulRead() throws Exception {
+    DocumentReference documentReference = randomColl.add(SINGLE_FIELD_MAP).get();
+
+    Timestamp firstWriteTime =
+        documentReference.set(Collections.singletonMap("counter", 1)).get().getUpdateTime();
+    documentReference.set(Collections.singletonMap("counter", 2)).get();
+
+    final TransactionOptions options =
+        TransactionOptions.createReadOnlyOptionsBuilder()
+            .setReadTime(
+                com.google.protobuf.Timestamp.newBuilder()
+                    .setSeconds(firstWriteTime.getSeconds())
+                    .setNanos(firstWriteTime.getNanos()))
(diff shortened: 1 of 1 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1490,6 +1490,41 @@ public void readOnlyTransaction_failureWhenAttemptingWrite()
...
+
+    final AtomicReference<DocumentSnapshot> ref = new AtomicReference<>();
+
+    final ApiFuture<Void> runTransaction =
+        firestore.runTransaction(
+            transaction -> {
+              final DocumentSnapshot snapshot =
+                  transaction.get(documentReference).get(5, TimeUnit.SECONDS);
+              ref.compareAndSet(null, snapshot);
+              return null;
```
schmidt-sebastian, line 1517: You could just return the value of the counter here. Then this transaction returns an `ApiFuture<Integer>` and you don't have to copy the document state.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -1490,6 +1490,41 @@ public void readOnlyTransaction_failureWhenAttemptingWrite()
...
 
+  @Test
+  public void readOnlyTransaction_successfulRead() throws Exception {
+    DocumentReference documentReference = randomColl.add(SINGLE_FIELD_MAP).get();
+
+    Timestamp firstWriteTime =
+        documentReference.set(Collections.singletonMap("counter", 1)).get().getUpdateTime();
+    documentReference.set(Collections.singletonMap("counter", 2)).get();
+
+    final TransactionOptions option =
```
schmidt-sebastian, line 1501: options?

schmidt-sebastian (commented):
This is great! Thank you for doing this.

One thing to read up on is https://www.conventionalcommits.org/en/v1.0.0/

The TL/DR is that the repo tooling using the PR title to generate releases. This means that your PR has to follow the prescribed pattern. You could use a name as such:

tests: add coverage for a successful read-only transaction with readTime

Note that this only applies to Java and Node.

schmidt-sebastian (approved):

PR #893, opened 2022-03-01
docs(sample): Use Java 8 for native image sample

mpeddada1:
This should fix the Native Image sample which is failing with the following error:

Screen Shot 2022-03-01 at 5 02 36 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,13 @@ implementation 'com.google.cloud:google-cloud-firestore'
 If you are using Gradle without BOM, add this to your dependencies
 
 ```Groovy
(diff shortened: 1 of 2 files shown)


Neenu1995 (approved):

suztomo (approved):
Can you also fix

```
              <mainClass>com.example.firestore.NativeImageFirestoreSample
              </mainClass>
```

to

```
              <mainClass>com.example.firestore.NativeImageFirestoreSample</mainClass>
```

suztomo:
I updated the title to reflect that the fix is not for library users.

PR #956, opened 2022-04-29
feat: add logical termination for RunQueryResponse

cherylEnkidu:
Add logical termination for RunQueryResponse.

When the SDK side receives either halfClose or RunQueryResponse.done set to true, the SDK will return user requested documents.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
@@ -1330,6 +1330,8 @@ public class Query {
 
     internalStream(
         new QuerySnapshotObserver() {
+          boolean hasCompleted = false;
+
           @Override
           public void onNext(QueryDocumentSnapshot documentSnapshot) {
             responseObserver.onNext(documentSnapshot);
@@ -1342,6 +1344,8 @@ public class Query {
 
           @Override
           public void onCompleted() {
+            if (hasCompleted) return;
+            hasCompleted = true;
             responseObserver.onCompleted();
           }
         },
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
@@ -1537,6 +1537,16 @@ public class Query {
             if (readTime == null) {
               readTime = Timestamp.fromProto(response.getReadTime());
             }
+
+            if (response.hasDone() && response.getDone()) {
+              Tracing.getTracer()
+                  .getCurrentSpan()
+                  .addAnnotation(
+                      "Firestore.Query: Completed",
+                      ImmutableMap.of(
+                          "numDocuments", AttributeValue.longAttributeValue(numDocuments)));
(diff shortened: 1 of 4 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1640,6 +1650,7 @@ ApiFuture<QuerySnapshot> get(@Nullable ByteString transactionId) {
     internalStream(
         new QuerySnapshotObserver() {
           final List<QueryDocumentSnapshot> documentSnapshots = new ArrayList<>();
+          boolean hasReturned = false;
```
wu-hui, line 1653: Maybe rename to `hasCompleted`.

Also add comment: // The stream's onCompleted could be called more than once, this flag makes sure only the first one is actually processed.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1655,6 +1666,8 @@ public void onError(Throwable throwable) {
           public void onCompleted() {
             // The results for limitToLast queries need to be flipped since we reversed the
             // ordering constraints before sending the query to the backend.
+            if (hasReturned) return;
```
wu-hui, line 1669: The comment should stay close to the limit type processing. Change it to:

```
if (hasCompleted) return;
hasCompleted = true;

// The results...
```

wu-hui (commented):
Requesting some minor changes.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/QueryTest.java
```diff
@@ -952,6 +953,76 @@ public void onCompleted() {
...
+            fail();
+          }
+
+          @Override
+          public void onCompleted() {
+            semaphore.release();
+          }
+        });
+
+    semaphore.acquire();
```
wu-hui, line 1023: We should verify onComplete is really only called once. Here it is not clear we are checking this. if `semaphore.release()` is called twice the test will still pass.

But the API does not give us a good way to verify it is not called a second time. I cannot think of a way to do this without introducing some timer.

What we can do is probably in `onCompleted()`, check to see if a flag is set from previous call, and throw if so. Then set that flag so we know it has been called.

This is not guaranteed to catch the problem, but at least we tried. We should also add a comment to explain why if you choose to do so.

wu-hui, line 1023: We should verify onComplete is really only called once. Here it is not clear we are checking this. if `semaphore.release()` is called twice the test will still pass.

But the API does not give us a good way to verify it is not called a second time. I cannot think of a way to do this without introducing some timer.

What we can do is probably in `onCompleted()`, check to see if a flag is set from previous call, and throw if so. Then set that flag so we know it has been called.

This is not guaranteed to catch the problem, but at least we tried. We should also add a comment to explain why if you choose to do so.

wu-hui (commented):
We are almost done here..

PR #1042, opened 2022-09-22
feat: Make count queries publicly available for use

dconeybe:
This PR changes the visibility of classes and methods that provide "count" queries from default to public, making them available for anyone to use. See #1033 for the implementation of COUNT queries.


ehsannas (approved):

ehsannas:
LGTM. Marked as "do not merge" for now.

PR #1044, opened 2022-09-22
test: Skip aggregateQueryInATransactionShouldRespectReadTime when run against the emulator

dconeybe:
The test aggregateQueryInATransactionShouldRespectReadTime fails when run against the Firestore emulator due to a known issue: b/220918135 (When creating a new read-only or read-write transaction, the Firestore emulator creates the transaction at the current time regardless of the user's specified TransactionsOptions. In the case of read-only transaction, if the user specifies the read_time, that will not be honoured. If the user specifies a retry_transaction, that will not be honored).

Instead of letting the test fail, this PR has it be skipped with an appropriate error message.


dconeybe:
It looks like b/220918135 is actually fixed. I've inquired to see if it will be released any time soon.

dconeybe:
cl/472994416, once submitted, will release a new version of the Firestore emulator with a fix for b/220918135.

wu-hui (approved):

PR #1050, opened 2022-09-30
test: Re-enable aggregateQueryInATransactionShouldRespectReadTime when run against the emulator

dconeybe:
This PR reverts https://github.com/googleapis/java-firestore/pull/1044, which disabled the test `aggregateQueryInATransactionShouldRespectReadTime` because of a known issue: b/220918135 (When creating a new read-only or read-write transaction, the Firestore emulator creates the transaction at the current time regardless of the user's specified TransactionsOptions. In the case of read-only transaction, if the user specifies the read_time, that will not be honoured. If the user specifies a retry_transaction, that will not be honored).

The issue has been confirmed fixed in `cloud-firestore-emulator-v1.15.1.jar`, so skipping the test is no longer required. This PR re-enables the test when run against he Firestore Emulator.


dconeybe:
Meh, I'm just going to merge this. Don't worry about the code review.

PR #1060, opened 2022-10-11
chore: Various improvements to COUNT queries

dconeybe:
1. Add an integration tests that uses an invalid collection name (i.e. `__invalid__`).


dconeybe:
Splitting into individual PRs.

PR #1062, opened 2022-10-12
feat: Add retry logic to COUNT queries

dconeybe:
When `AggregateQuery` was added by #1033 it failed to implement any "retry" logic. This PR adds in that retry logic, based heavily on the pre-existing retry logic from the `Query` class.


wu-hui (approved):
Sorry for the delay, LGTM!

PR #1076, opened 2022-10-24
chore: Use firestore-jar to determine grpc version

### LINKED ISSUE #1075: java.lang.NoClassDefFoundError: io/grpc/InternalMayRequireSpecificExecutor
We are getting this error when running tests in `samples/snippets`.
It is apparently a mismatch between compile and runtime versions of `grpc-core`
More details in the discussion of [this merged PR](https://github.com/googleapis/java-firestore/pull/1056#issuecomment-1277927134).

lqiu96:
[WIP] This is to test our theory about the GRPC jars

Fixes #1075 ☕️

### DIFF
--- a/samples/snapshot/pom.xml
+++ b/samples/snapshot/pom.xml
@@ -51,11 +51,47 @@
       <groupId>org.apache.beam</groupId>
       <artifactId>beam-runners-direct-java</artifactId>
       <version>${beam.version}</version>
+      <exclusions>
+        <exclusion>
+          <groupId>io.grpc</groupId>
+          <artifactId>grpc-api</artifactId>
+        </exclusion>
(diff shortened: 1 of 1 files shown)


lqiu96:
It seems the beam dependencies were pulling in the older versions of the `grpc-*` jars (v1.48.0 when it needed v1.49.2). The `firestore` dependency gets updated with the new `grpc-*` versions via shared-deps and we should have the it pull in the dependency going forward. 

I'll need figure out the warning with the snapshot jar being templated.

dconeybe (approved):
LGTM. Thanks for making this fix! The failing GitHub Action was such a nuisance. Plus, now the test coverage that the samples provide can be restored.

Issue #1152, opened 2022-12-19
Add support for serializing java record types


eranl:
Please add support for serializing java record types the same way POJOs are supported.

milaGGL:
Hi @eranl, happy holidays. 

Thank you for your suggestion. I will bring this up for team discussion and keep you updated.

eranl:
Hi @milaGGL,

Since record type support can't be added directly in this jdk8-based library, I have implemented some changes that allow for plugging in external custom mappers, such as for record types. Should I submit a PR with those changes?

I've also implemented the main code for a record mapper plugin, which could be released in a separate, jdk17-based library.

eranl:
Actually, there *is* a way to add record type support in this library, using reflection. I'm happy to implement such an approach as well.

milaGGL:
> Hi @milaGGL,
> 
> Since record type support can't be added directly in this jdk8-based library, I have implemented some changes that allow for plugging in external custom mappers, such as for record types. Should I submit a PR with those changes?
> 
> I've also implemented the main code for a record mapper plugin, which could be released in a separate, jdk17-based library.


Yes, please create a pull request. An informative PR will be very helpful when I bring this up to team discussion after the holiday break. 

If I understand correctly, the `record type` support needs to be implemented differently for jdk8-based library and jdk11-based library. Which library is using the `reflection`? Could you please create seperate PRs for both implementations?

eranl:
To clarify, the two options are:
1. add a record type mapper directly in this jdk8-based library, using reflection. Reflection is needed because records are not supported in jdk8.
2. add mapper plugin support to this library, and create a separate jdk17-based library containing a record type mapper, using that plugin support. Reflection would not be needed in this case, since records are fully supported in jdk17. This is the option I already have an implementation for.

Each approach has its advantages. option 2 would allow for plugging in any custom mapper, which can be seen as both an advantage and a disadvantage.

Let me submit a PR for option 2, to kick off  the discussion.

Happy New Year!

milaGGL:
Hi @eranl, just want to follow up on the PR you have mentioned.

eranl:
Hi @milaGGL,

PR posted.

milaGGL:
Thank you @eranl.  I will keep you updated on the discussion progress.

milaGGL:
Hi @ernal. After a team discussion on the PR, we are concerned about making the classes public. It is going to be a public API change in the SDK and there are very strict requirements for it. 

[...]

PR #1160, opened 2023-01-10
feat: Add support for plugging in custom bean mappers

eranl:
This is a proposed approach for enabling a fix for #1152. See discussion there for the different options.


eranl:
This approach has been deemed undesirable. See discussion at #1152.

PR #1267, opened 2023-04-24
feat: Add long alias support for aggregations.

ehsannas:


google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryAggregationsTest.java
```diff
@@ -81,6 +81,25 @@ public void canRunCountUsingAggregationMethod() throws Exception {
...
+    // The longest field name allowed is 1500. The alias chosen by the client is <op>_<fieldName>.
+    // If the field name is
+    // 1500 bytes, the alias will be longer than 1500, which is the limit for aliases. This is to
+    // make sure the client
+    // can handle this corner case correctly.
+    String longField = "";
+    for (int i = 0; i < 150; i++) {
+      longField += "0123456789";
+    }
+    Map<String, Map<String, Object>> testDocs = map("a", map(longField, 1), "b", map(longField, 2));
```
MarkDuckworth, line 95: I expected the field name to have the same 1500-byte limit as specified in document.proto. I'm unsure how this test works.

MarkDuckworth (commented):
LGTM other than the test that uses field names longer than 1500-bytes

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryAggregationsTest.java
MarkDuckworth: Sorry, that was an incomplete thought from me. According to [this string size calculation](https://firebase.google.com/docs/firestore/storage-size#document-id-size), this 1500 character string should result in a string length of 1501-bytes. This test may fail in production because the field size is too large.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -224,7 +222,11 @@ RunAggregationQueryRequest toProto(@Nullable final ByteString transactionId) {
       } else {
         throw new RuntimeException("Unsupported aggregation");
       }
-      aggregation.setAlias(aggregateField.getAlias());
+      // Map all client-side aliases to a unique short-form alias.
+      // This avoids issues with client-side aliases that exceed the 1500-byte string size limit.
+      String serverAlias = "aggregate_" + aggregationNum++;
+      aliasMap.put(serverAlias, aggregateField.getAlias());
+      aggregation.setAlias(serverAlias);
       aggregations.add(aggregation.build());
```
cherylEnkidu, line 230: will the deduplication still work?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryAggregationsTest.java
cherylEnkidu: Should we took 16 into consideration?
In your link example "6 + 5 + 6 + 11 + 16 = 44"

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryAggregationsTest.java
```diff
@@ -81,6 +81,25 @@ public void canRunCountUsingAggregationMethod() throws Exception {
...
+  public void allowsAliasesLongerThan1500Bytes() throws Exception {
+    // The longest field name allowed is 1500. The alias chosen by the client is <op>_<fieldName>.
+    // If the field name is
+    // 1500 bytes, the alias will be longer than 1500, which is the limit for aliases. This is to
+    // make sure the client
+    // can handle this corner case correctly.
+    String longField = "";
+    for (int i = 0; i < 150; i++) {
+      longField += "0123456789";
+    }
```
cherylEnkidu, line 94: Can we use StringBuilder just like this example?
```
StringBuilder builder = new StringBuilder(N);
for (int i = 0; i < N; i++) {
    builder.append("a");
}
String newString = builder.toString();
```

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryAggregationsTest.java
MarkDuckworth: I tested field name limits after posting my last comment. I found that the server did not add to the byte count when calculating the field name size. That is, in production, I was able to create a field name that was 1500-characters (and 1500 utf-8 encoded bytes).

Therefore, this test is good as is.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
ehsannas: My understanding is:
duplicate aliases *are not* allowed
duplicate aggregates *are* allowed

is that right @cherylEnkidu @MarkDuckworth ?

this code uses `aggregate_<counter>` as alias, so there should not be any duplicate aliases.

also, there's a test named `canGetDuplicateAggregations` below which currently passes against the emulator.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryAggregationsTest.java
ehsannas: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
ehsannas: ah... it's coming back to me now... I believe we decided to remove duplicate aggregates even though they are allowed to improve performance.

ehsannas: Done.

Followed what was done in Android. PTAL.

MarkDuckworth: LGTM

MarkDuckworth (approved):

ehsannas:
The CI is expected to fail (uses prod)

PR #1281, opened 2023-05-01
feat: Add toString methods to classes comprising WriteBatch

### LINKED ISSUE #1280: Add toString methods to classes comprising WriteBatch
For development, it would be very helpful to have toString methods on classes comprising WriteBatch. It would make it easier to see the data in the debugger and log it in dry-run mode.

eranl:
For development, it would be very helpful to have toString methods on classes comprising WriteBatch. It would make it easier to see the data in the debugger and log it in dry-run mode.

Fixes #1280 ☕️


tom-andersen:
Your change is very straight forward. I am inclined to approve, but would like a second from the team.

tom-andersen (changes requested):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
```diff
@@ -448,4 +448,10 @@ public boolean equals(Object obj) {
   public int hashCode() {
     return Objects.hash(rpcContext, docRef, fields);
   }
+
+  @Override
+  public String toString() {
+    return String.format("DocumentSnapshot{docRef=%s, fields=%s, readTime=%s, updateTime=%s, createTime=%s}", docRef,
```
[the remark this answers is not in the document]
eranl: Good points. Just committed. I used getSimpleName() instead, to avoid verbosity.

tom-andersen:
https://github.com/googleapis/java-firestore/blob/main/CONTRIBUTING.md

Code in this repo is formatted with [google-java-format](https://github.com/google/google-java-format). To run formatting on your project, you can run:

```
mvn com.coveo:fmt-maven-plugin:format
```

tom-andersen (changes requested):
Please add a unit test for each `toString` method.

You can create a new file in `src/test/java/com/google/cloud/firestore` for all `toString()` unit tests.

This is a slightly higher standard, since we haven't written unit tests for other `toString()` methods, so I hope you don't mind helping us improve our code quality. This also helps reviewers by providing an example of output.

I enabled CI on this PR. Whatever it complains about (ie code formatting), will need to be fixed before we can merge.

eranl:
Just committed take 3.
Sorry about forgetting to format the code. Done now.
I added a few unit tests. I hope that's what you had in mind. There's risk of high maintenance cost for such tests, so I'm not sure they're worth having.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
```diff
@@ -448,4 +448,11 @@ public boolean equals(Object obj) {
   public int hashCode() {
     return Objects.hash(rpcContext, docRef, fields);
   }
+
+  @Override
+  public String toString() {
+    return String.format(
+        "%s{docRef=%s, fields=%s, readTime=%s, updateTime=%s, createTime=%s}",
```
dconeybe, line 455: Here `docRef` is used, and in `UpdateBuilder.java` `documentReference` is used. How about just use `doc` consistently?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/ToStringTest.java
```diff
@@ -0,0 +1,136 @@
...
+            + "      string_value: \"value\"\n"
+            + "    }\n"
+            + "  }\n"
+            + "}\n"
+            + ", documentReference=DocumentReference{path=projects/test-project/databases/(default)/documents/coll/doc}}",
+        new UpdateBuilder.WriteOperation(documentReference, getDocumentSnapshot().toPb())
+            .toString());
+  }
+
+  private DocumentSnapshot getDocumentSnapshot() {
```
dconeybe, line 75: Please delete the `getDocumentSnapshot()` and instead inline it into the two test methods that use it. Although this results in a bit of code duplication it greatly improves the readability of the tests because that helper method defines key pieces of information that are necessary when constructing the expected result.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/ToStringTest.java
```diff
@@ -0,0 +1,136 @@
...
+    batch = firestoreMock.batch();
+    documentReference = firestoreMock.document("coll/doc");
+  }
+
+  @Test
+  public void testDocumentSnapshot() {
+    Assert.assertEquals(
+        "DocumentSnapshot{docRef=DocumentReference{path=projects/test-project/databases/(default)"
+            + "/documents/coll/doc}, fields={key=string_value: \"value\"\n"
+            + "}, readTime=null, updateTime=null, createTime=null}",
```
dconeybe, line 54: This test has a whole bunch of "nulls" in the toString() output. Could you instead use a DocumentSnapshot that uses non-null values? This will help ensure that the right text is associated with each key.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/ToStringTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+  @Before
+  public void before() {
+    batch = firestoreMock.batch();
+    documentReference = firestoreMock.document("coll/doc");
+  }
+
+  @Test
+  public void testDocumentSnapshot() {
+    Assert.assertEquals(
```
dconeybe, line 51: Testing the _exact_ return value of toString() is brittle (i.e. the test will fail with inconsequential changes to the return value of `toString()`; see https://abseil.io/resources/swe-book/html/ch12.html). Instead, please modify these tests to just check for "key information" in the string.

Also, please use `assertThat()` from `Truth` to do this, something like this, which IMO is more readable (note: untested code):

```java
String toStringResult = getDocumentSnapshot().toString();
assertThat(toStringResult).startsWith("DocumentSnapshot{");
assertThat(toStringResult).endsWith("}");
assertThat(toStringResult).contains("documents/coll/doc");
assertThat(toStringResult).contains("readTime=XXX");
```

Here is one other example from our test suite that uses a similar check: https://github.com/googleapis/java-firestore/blob/be5008f3de969df37932f8d844c8206597a59c14/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITQueryCountTest.java#L278-L282

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentSnapshot.java
eranl: These are just the field names, which happen to be different.

dconeybe: Right, but that implementation detail doesn't need to bleed out through the string returned from `toString()`. Since callers of `toString()` don't necessarily know, or even need to know, what the internal private variable name is, it's best to be consistent from that perspective.

eranl:
Just pushed take 4, which addresses @dconeybe's review comments, and limits verification to the toString implementation of the classes under test, as opposed to those of their fields.

tom-andersen:
Thanks @eranl.

Denver has been out of office, so there might be a delay before he has a chance to review.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/ToStringTest.java
```diff
@@ -0,0 +1,108 @@
...
+    documentReference = firestoreMock.document("coll/doc");
+  }
+
+  @Test
+  public void testDocumentSnapshot() {
+    Map<String, Value> fields =
+        Collections.singletonMap(
+            "key",
+            UserDataConverter.encodeValue(
+                FieldPath.of("key"),
```
dconeybe, line 58: nit: Please use more "distinct" strings than "key" and "value", such as "abc123" and "xyz987" or something like that. Doing this makes it easier to visually correlate the input to the DocumentSnapshot with its position in the string returned from toString(). Also, the strings "key" and "value" could show up in the toString() result in different contexts and be a bit confusing.

eranl:
Pushed take 5, addressing final comment

dconeybe (approved):
Thank you for bearing with us through the review process and for contributing to Firestore!

eranl:
Thank you @dconeybe for your thoughtful and timely review!
@tom-andersen, it looks like your approval is now required.

tom-andersen (approved):

PR #1323, opened 2023-06-10
fix: handling of superclass generic type parameters

### LINKED ISSUE #1313: Actual types passed to generic POJO super classes and to generic collection fields are ignored
Given the classes
```java
  private static class BaseBean<T> {
    public T value;
  }

  private static class SubBean extends BaseBean<String> {}
```
if you try to deserialize into the sub-bean, you get `java.lang.IllegalStateException: Could not resolve type T`. Same is true for a base bean with a setter that takes a T parameter.

While debugging this, I found that given the classes
```java
  private static class MapBean<T> {
    public Map<String, T> values;
  }

  private static class MapSubBean extends MapBean<String> {}

  private static class MapBeanHolder {
    public MapBean<String> map;
  }
```
deserializing into MapSubBean or MapBeanHolder, the code errs on the lenient side, and would accept any type of values into the `values` map.

I have a proposed fix for both issues. Should I post a PR? The collection part of the fix might break existing code, so it may need to be excluded.

eranl:
Enhance generic type resolution, so that no actual type arguments are lost.

For backward compatibility, current lenient behavior of generic collection fields is preserved, unless the new `@StrictCollectionTypes` annotation is used. A few of the new unit tests verify this lenient behavior. These tests pass with and without this change.

Fixes #1313 ☕️

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
@@ -22,6 +22,7 @@ import com.google.cloud.firestore.annotation.Exclude;
 import com.google.cloud.firestore.annotation.IgnoreExtraProperties;
 import com.google.cloud.firestore.annotation.PropertyName;
 import com.google.cloud.firestore.annotation.ServerTimestamp;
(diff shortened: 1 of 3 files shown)


eranl:
@tom-andersen, @milaGGL, can you please take a look at this PR? I'm happy to rebase it onto current main if you are ready for it.

PR #1350, opened 2023-07-07
Fix javadoc failure

milaGGL:
Adding / updating comments to pass javadoc lint check

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
@@ -53,7 +53,7 @@ public class AggregateQuery {
     this.query = query;
   }
 
-  /** Returns the query whose aggregations will be calculated by this object. */
+  /** @return The query whose aggregations will be calculated by this object. */
   @Nonnull
   public Query getQuery() {
     return query;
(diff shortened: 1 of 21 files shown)


suztomo:
Assuming you're working on javadoc error fix, you want to focus on the "errors: " lines rather than "warnings: ". Note that the warnings for javadoc may not match Google's coding convention. In that case you don't have to fix warnings.

> Tip: A common mistake is to write simple Javadoc in the form /** @return the customer ID */. This is incorrect, and should be changed to /** Returns the customer ID. */.

https://google.github.io/styleguide/javaguide.html

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -53,7 +53,7 @@ public class AggregateQuery {
     this.query = query;
   }
 
-  /** Returns the query whose aggregations will be calculated by this object. */
+  /** @return the query whose aggregations will be calculated by this object. */
```
tom-andersen, line 56: The sentence starts after the `@return` annotation, so make the first letter a capital. Please do this for other texts as well.

```suggestion
  /** @return The query whose aggregations will be calculated by this object. */
```

suztomo:
You want to focus on the "errors: " lines rather than "warnings: ".

milaGGL:
Re-opening a new PR which only addresses errors, not warnings.

milaGGL:
> You want to focus on the "errors: " lines rather than "warnings: ".

Thank you @suztomo, javadoc test passed after fixing the "errors".

PR #1374, opened 2023-07-25
Chore: Setup integration tests against named db

wu-hui:


tom-andersen (approved):
There is some Kokoro magic I don't understand. But the Java changes look good!

PR #1499, opened 2023-12-07
test: prime watch before running tests.

ehsannas:
This should hopefully help eliminate some of the flakiness we see, such as: https://github.com/googleapis/java-firestore/issues?page=1&q=is%3Aissue+label%3A%22flakybot%3A+issue%22+ITQueryWatchTest

This was ported from https://github.com/firebase/firebase-android-sdk/blob/080e1fcd3a3632fd21b49d8cb4076bc0d34d2904/firebase-firestore/src/androidTest/java/com/google/firebase/firestore/testutil/IntegrationTestUtil.java#L212-L246

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITBaseTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITBaseTest.java
@@ -35,7 +35,7 @@ public abstract class ITBaseTest {
   private FirestoreOptions firestoreOptions;
 
   @Before
-  public void before() {
+  public void before() throws Exception {
     FirestoreOptions.Builder optionsBuilder = FirestoreOptions.newBuilder();
 
     String namedDb = System.getProperty("FIRESTORE_NAMED_DATABASE");
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITBulkWriterTest.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITBulkWriterTest.java
@@ -48,7 +48,7 @@ public class ITBulkWriterTest extends ITBaseTest {
   @Rule public TestName testName = new TestName();
 
   @Before
-  public void before() {
+  public void before() throws Exception {
     super.before();
     randomColl =
         firestore.collection(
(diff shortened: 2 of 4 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITBaseTest.java
```diff
@@ -48,6 +65,45 @@ public void before() {
 
     firestoreOptions = optionsBuilder.build();
     firestore = firestoreOptions.getService();
+    primeBackend();
+  }
+
+  public void primeBackend() throws Exception {
+    if (!backendPrimed) {
```
tom-andersen, line 72: I think you should refactor this to be a guard thereby flattening the indentation. It also makes reading easier because you don't have to search for an `else` code block.

```
if (backendPrimed) return;
```

This is one of those contentious situations where I think a one liner without curly braces is acceptable, but I doubt that follows the style guide.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITBaseTest.java
```diff
@@ -48,6 +65,45 @@ public void before() {
...
+      backendPrimed = true;
+      CompletableFuture<Void> watchInitialized = new CompletableFuture<>();
+      CompletableFuture<Void> watchUpdateReceived = new CompletableFuture<>();
+      DocumentReference docRef = firestore.collection(autoId()).document();
+      ListenerRegistration listenerRegistration =
+          docRef.addSnapshotListener(
+              (snapshot, error) -> {
+                assertNull(error);
```
tom-andersen, line 80: The assert is swallowed. If error is ever not null, nothing is propagated.

I don't know how best to handle this. There are better testing frameworks than vanila JUnit4 for asynchronous testing. I like when the test method receives a future `primeBackend(future)` that can be called asynchronously to indicate success or failure. That would make this a lot easier.

A rubber bandaid solution would be to do something like the following:

```
if (error != null) {
  logger.log(Level.SEVERE, "Prime backend received error in snapshot listener.", error);
  if (!watchInitialized.isDone()) {
    watchInitialized.completeExceptionally(error);
  } else if (!watchUpdateReceived.isDone()) {
    watchUpdateReceived.completeExceptionally(error);
  }
}
```

tom-andersen (changes requested):

tom-andersen (approved):
LGTM

PR #1506, opened 2023-12-14
test: Increase Query test coverage

cherylEnkidu:


google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -368,13 +368,107 @@ public void defaultQuery() throws Exception {
...
+
+          @Override
+          public void onCompleted() {
+            semaphore.release();
+          }
+        });
+
+    semaphore.acquire();
+    // Verify the number of response
+    assertFalse(iterator.hasNext());
```
milaGGL, line 396: Assert document count instead of asserting  iterator emptiness would be clearer.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -368,13 +368,107 @@ public void defaultQuery() throws Exception {
...
+    final int[] docCount = {0};
+    randomColl.stream(
+        new ApiStreamObserver<DocumentSnapshot>() {
+          @Override
+          public void onNext(DocumentSnapshot documentSnapshot) {
+            docCount[0]++;
+          }
+
+          @Override
+          public void onError(Throwable throwable) {}
```
milaGGL, line 427: Consider adding assertions within the onError callback to catch any unexpected exceptions

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
```diff
@@ -368,13 +368,107 @@ public void defaultQuery() throws Exception {
     assertEquals("bar", documents.next().get("foo"));
   }
 
+  @Test
+  public void defaultQueryStream() throws Exception {
+    addDocument("foo", "bar");
```
milaGGL, line 373: I would prefer having different field values to make sure each DocumentSnapshot we are getting is different.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITSystemTest.java
cherylEnkidu: `assertFalse(iterator.hasNext());` is also used in other existing test functions inside this file. IMO count is less intuitive than iterator since in Java test count needs to wrapped inside an array and introduces one more variable  to this test where the same purpose can be achieved by iterator.

milaGGL (approved):
LGTM.

PR #1515, opened 2024-01-04
fix: Make rollback best effort.

tom-andersen:
Internal tracking: b/318686162

A rollback is meant to be best effort. If the transaction has already expired, it is possible for the rollback to fail due to transaction no longer existing in Firestore. The retry logic will use attempts to rollback, and in the case where transaction no longer exists, all attempts to be exhausted attempted to rollback transaction.

This PR makes the rollback best effort, simply logging any rollback error and continuing.

### DIFF
--- a/google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
+++ b/google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
@@ -1196,10 +1196,6 @@ public final class LocalFirestoreHelper {
         Answer<ApiFuture<? extends GeneratedMessageV3>> answer =
             invocationOnMock -> {
               actualRequestList.add(invocationOnMock.getArguments()[0]);
-              if (!entry.request.equals(invocationOnMock.getArguments()[0])) {
-                System.out.println("BAD INVOCATION");
-                System.out.println(invocationOnMock.getArguments()[0]);
-              }
               return entry.response;
             };
         stubber = (stubber != null) ? stubber.doAnswer(answer) : doAnswer(answer);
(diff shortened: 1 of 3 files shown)


google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
```diff
@@ -1193,8 +1195,11 @@ void initializeStub(
       for (final RequestResponsePair entry : operationList) {
         Answer<ApiFuture<? extends GeneratedMessageV3>> answer =
             invocationOnMock -> {
-              ++requestCount;
-              assertEquals(entry.request, invocationOnMock.getArguments()[0]);
+              actualRequestList.add(invocationOnMock.getArguments()[0]);
+              if (!entry.request.equals(invocationOnMock.getArguments()[0])) {
+                System.out.println("BAD INVOCATION");
+                System.out.println(invocationOnMock.getArguments()[0]);
```
ehsannas, line 1201: remove system.out

ehsannas (commented):
Even though the documentation suggests that rollbacks are best-effort, we may want to still keep up the effort to retry the rollback unless we're positive the retry will fail (i.e. for INTERNAL error code).

google-cloud-firestore/src/test/java/com/google/cloud/firestore/LocalFirestoreHelper.java
tom-andersen: Removed

tom-andersen:
> Even though the documentation suggests that rollbacks are best-effort, we may want to still keep up the effort to retry the rollback unless we're positive the retry will fail (i.e. for INTERNAL error code).

I verified that GAPIC does do the retries from `rollback` appropriately following the service config. So we can safely remove this extra layer that was simply retying errors that couldn't be retried anyway.

ehsannas (approved):

PR #1523, opened 2024-01-10
fix: `internalStream` should handle duplicate `onComplete`s.

ehsannas:
The burden of handling duplicate `onComplete` signals should be on `internalStream`, and not on all the callers that use it.

currently `get()` and `stream()` APIs call the underlying `internalStream`, and each of them are implementing the same logic to avoid duplicate completions. This should be handled at the lower (internalStream) function. Also, future callers of `internalStream` may not realize they need to implement this logic.


cherylEnkidu (approved):
Thank you!

PR #1536, opened 2024-01-19
refactor: Optimize FieldMask instantiation

tom-andersen:
Optimization does 2 things:

1. Avoid creating intermediary collections. Simply create a TreeSet that is used in FieldMask constructor.
2. Instead of adding FieldPaths to an intermediary collection and then removing them afterwards, simply filter FieldPaths as part of TreeSet creation.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
@@ -19,6 +19,7 @@ package com.google.cloud.firestore;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.SortedSet;
@@ -47,7 +48,7 @@ public final class FieldMask {
    */
   @Nonnull
   public static FieldMask of(String... fieldPaths) {
-    List<FieldPath> paths = new ArrayList<>();
+    TreeSet<FieldPath> paths = new TreeSet<>();
     for (String fieldPath : fieldPaths) {
       paths.add(FieldPath.fromDotSeparatedString(fieldPath));
     }
@@ -62,16 +63,18 @@ public final class FieldMask {
    */
   @Nonnull
   public static FieldMask of(FieldPath... fieldPaths) {
-    return new FieldMask(Arrays.asList(fieldPaths));
+    TreeSet<FieldPath> paths = new TreeSet<>();
+    Collections.addAll(paths, fieldPaths);
+    return new FieldMask(paths);
   }
 
   static FieldMask fromObject(Map<String, Object> values) {
-    List<FieldPath> fieldPaths = extractFromMap(values, FieldPath.empty());
+    TreeSet<FieldPath> fieldPaths = extractFromMap(values, FieldPath.empty());
     return new FieldMask(fieldPaths);
   }
 
-  private static List<FieldPath> extractFromMap(Map<String, Object> values, FieldPath path) {
-    List<FieldPath> fieldPaths = new ArrayList<>();
+  private static TreeSet<FieldPath> extractFromMap(Map<String, Object> values, FieldPath path) {
+    TreeSet<FieldPath> fieldPaths = new TreeSet<>();
 
     for (Map.Entry<String, Object> entry : values.entrySet()) {
       Object value = entry.getValue();
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
@@ -16,11 +16,8 @@
 
 package com.google.cloud.firestore;
 
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.List;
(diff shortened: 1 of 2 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
```diff
@@ -47,7 +45,7 @@ private FieldMask(SortedSet<FieldPath> fieldPaths) {
    */
   @Nonnull
   public static FieldMask of(String... fieldPaths) {
-    List<FieldPath> paths = new ArrayList<>();
+    TreeSet<FieldPath> paths = new TreeSet<>();
```
dconeybe, line 48: The change from ArrayList to TreeSet in this class looks like strictly _worse_ performance, since adding to a TreeSet is slower than adding to an ArrayList. Also, using TreeSet subtly changes the semantics because a TreeSet will eliminate duplicates and potentially change the order of elements where an ArrayList would keep them. Could you explain the motivation for changing to TreeSet?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
dconeybe, line 1: Do you have any statistics to show how much of a performance gain is realized here, and under what conditions? I ask because it's not clear to me that there will be a meaningful performance gain since TreeSet comes with a peformance penalty due to how it's implemented. TreeSet also detracts from readability IMO because it is a collection that's only used in particular cases and its use immediately raises the question in reader's minds "why it is chosen rather than a hash-based or array-based structure like HashSet or ArrayList?". So I'd discourage this change unless there is some proven performance gain.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
```diff
@@ -62,16 +60,18 @@ public static FieldMask of(String... fieldPaths) {
...
 
   static FieldMask fromObject(Map<String, Object> values) {
-    List<FieldPath> fieldPaths = extractFromMap(values, FieldPath.empty());
+    TreeSet<FieldPath> fieldPaths = extractFromMap(values, FieldPath.empty());
     return new FieldMask(fieldPaths);
   }
 
-  private static List<FieldPath> extractFromMap(Map<String, Object> values, FieldPath path) {
-    List<FieldPath> fieldPaths = new ArrayList<>();
+  private static TreeSet<FieldPath> extractFromMap(Map<String, Object> values, FieldPath path) {
```
dconeybe, line 73: A potentially better optimization here would be to modify the `extractFromMap()` method to take the destination collection as an argument, instead of creating a new collection on every invocation. This is especially true since this method calls itself recursively and creates a new collection on each invocation.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -272,9 +275,11 @@ private T performSet(
         DocumentTransform.fromFieldPathMap(documentReference, documentData);
 
     if (options.getFieldMask() != null) {
-      List<FieldPath> fieldMask = new ArrayList<>(options.getFieldMask());
-      fieldMask.removeAll(documentTransform.getFields());
-      documentMask = new FieldMask(fieldMask);
+      TreeSet<FieldPath> fieldPaths =
```
dconeybe, line 278: Why is TreeSet used in this class too instead of ArrayList? I agree that the streaming API is a good improvement, but I'm not clear on the benefits of TreeSet, which could negate any performance gains.

dconeybe (changes requested):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
dconeybe: Nevermind. @tom-andersen pointed out to me that TreeSet was already there; this change avoids creating an intermediary collection that is (almost) immediately converted to a TreeSet then thrown away.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
dconeybe: Nevermind. @tom-andersen pointed out to me that TreeSet was already there; this change avoids creating an intermediary collection that is (almost) immediately converted to a TreeSet then thrown away.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FieldMask.java
dconeybe: Nevermind. @tom-andersen pointed out to me that TreeSet was already there; this change avoids creating an intermediary collection that is (almost) immediately converted to a TreeSet then thrown away.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
dconeybe: Nevermind. @tom-andersen pointed out to me that TreeSet was already there; this change avoids creating an intermediary collection that is (almost) immediately converted to a TreeSet then thrown away.

dconeybe (approved):
LGTM. Thanks for this improvement, and for gracefully pointing out my uninformed review comments!

PR #1537, opened 2024-01-19
fix: Thread safe UpdateBuilder

tom-andersen:
b/318413194

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
@@ -69,6 +69,8 @@ class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
    * <p>The writes in the batch are not applied atomically and can be applied out of order.
    */
   ApiFuture<Void> bulkCommit() {
+    committed = true;
+
     BatchWriteRequest request = buildBatchWriteRequest();
 
     Tracing.getTracer()
@@ -77,8 +79,6 @@ class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
             TraceUtil.SPAN_NAME_BATCHWRITE,
             ImmutableMap.of(
                 "numDocuments", AttributeValue.longAttributeValue(request.getWritesCount())));
-
-    committed = true;
 
     ApiFuture<BatchWriteResponse> response =
         processExceptions(
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
@@ -69,24 +69,20 @@ class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
    * <p>The writes in the batch are not applied atomically and can be applied out of order.
    */
   ApiFuture<Void> bulkCommit() {
+    BatchWriteRequest request = buildBatchWriteRequest();
+
     Tracing.getTracer()
         .getCurrentSpan()
         .addAnnotation(
             TraceUtil.SPAN_NAME_BATCHWRITE,
-            ImmutableMap.of("numDocuments", AttributeValue.longAttributeValue(getWrites().size())));
-
-    final BatchWriteRequest.Builder request = BatchWriteRequest.newBuilder();
-    request.setDatabase(firestore.getDatabaseName());
-
-    for (WriteOperation writeOperation : getWrites()) {
-      request.addWrites(writeOperation.write);
-    }
+            ImmutableMap.of(
+                "numDocuments", AttributeValue.longAttributeValue(request.getWritesCount())));
 
     committed = true;
 
     ApiFuture<BatchWriteResponse> response =
         processExceptions(
-            firestore.sendRequest(request.build(), firestore.getClient().batchWriteCallable()));
+            firestore.sendRequest(request, firestore.getClient().batchWriteCallable()));
 
     return ApiFutures.transformAsync(
         response,
@@ -115,6 +111,13 @@ class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
           return BulkWriter.silenceFuture(ApiFutures.allAsList(pendingUserCallbacks));
         },
         executor);
+  }
+
+  private BatchWriteRequest buildBatchWriteRequest() {
+    BatchWriteRequest.Builder builder = BatchWriteRequest.newBuilder();
+    builder.setDatabase(firestore.getDatabaseName());
+    forEach(builder::addWrites);
+    return builder.build();
   }
 
   /** Maps an RPC failure to each individual write's result. */
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
@@ -88,18 +88,18 @@ class BulkCommitBatch extends UpdateBuilder<ApiFuture<WriteResult>> {
     return ApiFutures.transformAsync(
         response,
         batchWriteResponse -> {
+          List<ApiFuture<Void>> pendingUserCallbacks = new ArrayList<>();
+
           List<com.google.firestore.v1.WriteResult> writeResults =
               batchWriteResponse.getWriteResultsList();
           List<com.google.rpc.Status> statuses = batchWriteResponse.getStatusList();
 
-          int size = writeResults.size();
-          List<ApiFuture<Void>> pendingUserCallbacks = new ArrayList<>(size);
-          for (int i = 0; i < size; ++i) {
+          for (int i = 0; i < writeResults.size(); ++i) {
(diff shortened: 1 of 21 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -62,7 +63,7 @@ public String toString() {
 
   final FirestoreImpl firestore;
 
-  private final List<WriteOperation> writes = new ArrayList<>();
+  private final List<WriteOperation> writes = new CopyOnWriteArrayList<>();
```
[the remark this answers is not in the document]
tom-andersen: I added comment indicating need for a thread safe collection.

I did not add a comment as to why CopyOnWriteArrayList was chosen. There are alternatives.

I followed advice from [go/java-practices/collection-implementations](https://goto.google.com/java-practices/collection-implementations).

Collections.synchronizedList(..) wrapper could also have been used.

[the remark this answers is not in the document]
tom-andersen: Note use of `Vector` is discouraged.

go/java-practices/choosing-collections#avoid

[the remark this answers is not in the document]
dconeybe: Thanks for that link. Still, `CopyOnWriteArrayList` doesn't feel like the correct data structure here as it is optimized for the use case where reads dominate writes; however, the usage pattern here seems to be the exact opposite. This is mainly because appending an element to a `CopyOnWriteArrayList` is an O(n) operation (compared to ArrayList which is O(1) on average). So adding n elements to a `CopyOnWriteArrayList` is an O(n^2) operation. Can you confirm that this performance is acceptable?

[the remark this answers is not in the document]
dconeybe: Also, changing it to a `CopyOnWriteArrayList` still leaves some race conditions in the code. Namely, there are several places that access the `writes` array multiple times seemingly under the assumption that it will not undergo any modifications between those accesses.

For example, this pattern is used often:

```java
writes.add(...);
return wrapResult(writes.size() - 1);
```

This is not thread-safe because the size of the `writes` array could change between the calls to `writes.add()` and `writes.size()` in the presence of concurrent modifications to `write`.

Also, in `commit()` it logs `writes.size()` then iterates over it in a separate step. This could be really confusing if there is an intervening modification to `writes` and the logged number of writes differs from the _actual_ number of writes that were performed.

Also, the `committed` member variable is another source of a data race.

Basically, to make this class truly thread safe will require more work than just swapping out ArrayList for a "thread-safe" list.

dconeybe (changes requested):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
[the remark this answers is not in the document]
tom-andersen: Thank you for identifying concurrency problems.

Intent of original PR was to prevent corruption of `ArrayList` during proper usage.

Improper usage of calling write after commit wasn't really considered in scope. There is a check that already throws an exception telling developer not to do this. But we can strengthen this, to also throw an exception for concurrent case that can intermittently pass precondition.

I don't think the `writes.size() - 1` is a real problem because only the BulkWriter uses this and already has thread safety. But for the sake of completeness, I will add `synchronization` to UpdateBuilder as a second layer of safety and protection of future development.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -48,13 +52,13 @@
  */
 @InternalExtensionOnly
 public abstract class UpdateBuilder<T> {
-  static class WriteOperation {
-    Write.Builder write;
-    DocumentReference documentReference;
+  static final class WriteOperation {
```
tom-andersen, line 55: This class is easier to reason about if immutable. Final also gives slight performance improvement.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
```diff
@@ -117,6 +114,13 @@ ApiFuture<Void> bulkCommit() {
         executor);
   }
 
+  private BatchWriteRequest buildBatchWriteRequest() {
```
dconeybe, line 117: nit: I think this method does not need to exist, and can just be inlined at the single call site (where it was previously inlined).

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
```diff
@@ -69,40 +69,37 @@ ApiFuture<WriteResult> wrapResult(int writeIndex) {
...
         batchWriteResponse -> {
-          List<ApiFuture<Void>> pendingUserCallbacks = new ArrayList<>();
-
           List<com.google.firestore.v1.WriteResult> writeResults =
               batchWriteResponse.getWriteResultsList();
           List<com.google.rpc.Status> statuses = batchWriteResponse.getStatusList();
 
-          for (int i = 0; i < writeResults.size(); ++i) {
-            com.google.firestore.v1.WriteResult writeResult = writeResults.get(i);
+          int size = writeResults.size();
```
dconeybe, line 95: It doesn't look like the changes in `transformAsync()` are fixing any concurrency bugs. Can these changes be reverted in this PR? Fully addressing the concurrency bugs in `BulkCommitBatch` is for another PR IMO.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -293,13 +282,25 @@ private T performSet(
       write.setUpdateMask(documentMask.toPb());
     }
 
-    writes.add(new WriteOperation(documentReference, write));
+    return writesAdd(new WriteOperation(documentReference, write));
+  }
 
-    return wrapResult(writes.size() - 1);
+  private T writesAdd(WriteOperation operation) {
```
dconeybe, line 288: nit: Consider renaming `writesAdd()` to `addWrite()`, which, IMO, is a more conventional name. Also, consider having this method take the `DocumentReference` and `Write` as an argument to deduplicate creating of the `WriteOperation` at the call sites.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
```diff
@@ -595,76 +594,98 @@ public T delete(@Nonnull DocumentReference documentReference) {
...
   boolean isEmpty() {
-    return writes.isEmpty();
+    synchronized (writes) {
+      return writes.isEmpty();
+    }
   }
 
-  List<WriteOperation> getWrites() {
-    return writes;
+  void forEach(Consumer<Write> consumer) {
```
dconeybe, line 666: nit: Consider renaming `forEach` to `forEachWrite` or something like that. Personally, I just find that `forEach` does not immediately convey _what_ it is iterating over.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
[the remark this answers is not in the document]
dconeybe: After discussing in person, changes were made that satisfy these concerns.

dconeybe (changes requested):
Thanks for making this improvement! I've just left a few minor nits, but overall LGTM.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
tom-andersen: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/UpdateBuilder.java
tom-andersen: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
tom-andersen: There is some value in separating out request builder logic.

The `builder` variable is encapsulated in the scope of method, and any interaction with built request is through an immutable object. This lowers the mental burden of reasoning about `commit()`.

Also breaking larger code blocks down into smaller more easily digestible blocks helps readability. go/java-practices/methods#length

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
tom-andersen: I'll revert.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/BulkCommitBatch.java
dconeybe: Ok sounds good.

dconeybe (approved):

PR #1565, opened 2024-02-08
fix: Transactions with `readTime` will omit begin and commit transaction requests, and instead pass `readTime` on individual read requests.

tom-andersen:
Transactions with `readTime` do not need to begin and commit transaction. Instead, the individual requests  can use `readTime` instead of `transactionId`. This will reduce latency when doing point in time reads.

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -29,6 +29,7 @@ import com.google.api.gax.rpc.ServerStreamingCallable;
 import com.google.api.gax.rpc.StreamController;
 import com.google.api.gax.rpc.UnaryCallable;
 import com.google.cloud.Timestamp;
+import com.google.cloud.firestore.TransactionOptions.TransactionOptionsType;
 import com.google.cloud.firestore.spi.v1.FirestoreRpc;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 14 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -194,28 +194,29 @@ public Iterable<CollectionReference> listCollections() {
   @Override
   public ApiFuture<List<DocumentSnapshot>> getAll(
       @Nonnull DocumentReference... documentReferences) {
-    return this.getAll(documentReferences, null, (ByteString) null);
+    return this.getAll(documentReferences, null, (ByteString) null, null);
```
milaGGL, line 197: "getAll" functions are called with too many "null"s. Maybe make a overload function call the next overload function that has just one more parameter, and chain them up. e.g, here we call:
`return this.getAll(documentReferences, null);`, 
which should call
`public ApiFuture<List<DocumentSnapshot>> getAll(
      @Nonnull DocumentReference[] documentReferences, @Nullable FieldMask fieldMask) `

google-cloud-firestore/src/main/java/com/google/cloud/firestore/ServerSideTransactionRunner.java
```diff
@@ -202,7 +205,7 @@ private ApiFuture<T> restartTransactionCallback(Throwable throwable) {
     }
 
     ApiException apiException = (ApiException) throwable;
-    if (transaction.hasTransactionId() && isRetryableTransactionError(apiException)) {
```
milaGGL, line 205: would remove the transaction Id existence check cause any problems?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/ServerSideTransaction.java
```diff
@@ -0,0 +1,208 @@
...
+import com.google.protobuf.ByteString;
+import com.google.protobuf.Empty;
+import io.opencensus.trace.Tracing;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+final class ServerSideTransaction extends Transaction {
```
milaGGL, line 37: what are the main difference between `ServerSideTransaction` and `ReadTimeTransaction`? can we add comments for these 2 classes for better understanding.

tom-andersen (commented):
Thanks for the feedback. I think I have addressed all points, and pushed up changes.

milaGGL (approved):
LGTM, with one quick question. readTime is added for Query and AggregateQuery, what about single documents?

tom-andersen:
> one quick question. readTime is added for Query and AggregateQuery, what about single documents?

The proto doesn't support transaction or readTime on single get. Whenever transaction or readTime is required, we use **batch** get proto, even if it is for a single document.

PR #1590, opened 2024-02-27
feat: tracing for aggregate queries, bulkwriter, partition queries, a…

ehsannas:


google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
```diff
@@ -102,21 +104,32 @@ public ApiFuture<List<QueryPartition>> getPartitions(long desiredPartitionCount)
...
+                        partitions.add(queryPartition);
+                        return null;
+                      });
+                  return partitions.build();
+                },
+                MoreExecutors.directExecutor());
+        span.endAtFuture(result);
+        return result;
       } catch (ApiException exception) {
+        span.end(exception);
```
cindy-peng, line 132: In openCensus, we do `span.end()` in the `finally` block:https://github.com/googleapis/java-firestore/pull/1589/files#diff-04b7fd9ffd834a763fe56dd88c3e16d42bd8d1d3e3df4b134ae9ad296f147c00L133

Will there be an edge case where exception thrown but not `ApiException` type, and we end up failing to end the span?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
```diff
@@ -127,48 +129,56 @@ public DocumentReference document(@Nonnull String childPath) {
...
+                }
+
+                @Override
+                public void remove() {
+                  throw new UnsupportedOperationException("remove");
+                }
+              };
+            }
+          };
+      span.end();
```
cindy-peng, line 176: Do we need `endAtFuture` here?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
```diff
@@ -0,0 +1,304 @@
+/*
+ * Copyright 2023 Google LLC
```
cindy-peng, line 2: Nitpick: 2024?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionGroup.java
ehsannas: yeah, I think that's a legitimate corner case. There's a subtle difference with OpenCensus -- which is that with open-telemetry we end the span differently based on whether there was an exception or not. So I will add a `catch` block to catch all other kinds of Throwables, instead of adding a `finally` block.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
```diff
@@ -0,0 +1,304 @@
...
+      for (SpanData spanData : spans) {
+        assertEquals(traceId, spanData.getTraceId());
+      }
+    }
+  }
+
+  // Helper to see the spans in standard output while developing tests
+  void printSpans() {
+    for (SpanData spanData : spanNameToSpanData.values()) {
+      System.out.printf(
```
kolea2, line 178: maybe switch this to a logging library with levels? That way we don't clutter up the logs when running the ITs

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
```diff
@@ -0,0 +1,304 @@
...
+  private static final String LIST_DOCUMENTS_RPC_NAME = "ListDocuments";
+  private static final String BATCH_WRITE_RPC_NAME = "BatchWrite";
+
+  // We use an InMemorySpanExporter for testing which keeps all generated trace spans
+  // in memory so that we can check their correctness.
+  protected static InMemorySpanExporter inMemorySpanExporter = InMemorySpanExporter.create();
+
+  protected Firestore firestore;
+
+  HashMap<String, String> spanNameToSpanId = new HashMap<>();;
```
kolea2, line 71: nit - switch to `Map` on lefthand side.

kolea2 (changes requested):

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
kolea2: and remove double `;;` to `;`

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
ehsannas: Done.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
ehsannas: Done.

cindy-peng (approved):

kolea2 (approved):

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
```diff
@@ -0,0 +1,306 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore.it;
+
+import static com.google.cloud.firestore.telemetry.TraceUtil.*;
+import static io.opentelemetry.semconv.resource.attributes.ResourceAttributes.SERVICE_NAME;
+import static org.junit.Assert.*;
+
+import com.google.cloud.firestore.*;
```
kolea2, line 23: nit - use single class imports over wildcards

kolea2 (commented):
lgtm with one nit!

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
ehsannas: Done. Thanks for pointing these out. [I swear this is my IDE adding imports behind my back as I'm coding 🤦‍♂️ ]

PR #1591, opened 2024-02-27
feat: trace instrumentation for DocumentReference methods.

ehsannas:


google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
```diff
@@ -372,41 +513,45 @@ public ApiFuture<DocumentSnapshot> get(FieldMask fieldMask) {
...
+                }
+              };
+            }
+          };
+      span.end();
+      return result;
     } catch (ApiException exception) {
+      span.end(exception);
       throw FirestoreException.forApiException(exception);
     }
```
kolea2, line 554: maybe this is the diff, but why is the order changing here (catching exceptions then returning vs a catch all the way at the end)?

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
```diff
@@ -59,7 +59,10 @@ public class ITTracingTest {
       Logger.getLogger(com.google.cloud.firestore.it.ITBaseTest.class.getName());
 
   private static final String SERVICE = "google.firestore.v1.Firestore/";
+  private static final String BATCH_GET_DOCUMENTS_RPC_NAME = "BatchGetDocuments";
```
kolea2, line 62: is this used in the actual code as well? May make sense to make it a protected variable so you don't have to copy over here (same goes for others)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/DocumentReference.java
ehsannas: I found the way the code was originally written hard to read: the `try` block used to only encapsulate part of the logic. I decided to improve this a bit by moving the `return` statements into the `try` block.

This shouldn't change the behavior because the only exception that could be thrown by the return statement is an "out of memory" exception (because of the `new`). And in the case of OOM, whether the return statement is in the `try` block or after it, the behavior would be the same (because the function doesn't catch any OOM exceptions). On the plus side though, by moving the return statement into the try block, we can make the "SPAN_NAME_DOC_REF_LIST_COLLECTIONS span" nicely encapsulate all the cpu time spent in this function.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITTracingTest.java
ehsannas: These tests ensure that 2 kinds of trace spans are generated:
1- spans generated in the Firestore handwritten layer.
2- spans generated by the gRPC auto-instrumentation layer.

For #1: we do reuse variables from our `TraceUtil.java` code.
For #2: the spans correspond to the gRPC request created in by the GAPIC layer, and there doesn't seem to be a variable defined that we can reuse here. See: https://github.com/googleapis/java-firestore/blob/532e094b1a428af2f03fd91202e718a428366b5a/grpc-google-cloud-firestore-v1/src/main/java/com/google/firestore/v1/FirestoreGrpc.java#L238

kolea2 (approved):

PR #1609, opened 2024-03-13
feat: Query Profiling

ehsannas:


google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -136,28 +213,45 @@ private final class AggregateQueryResponseObserver
...
-      // if it is called more than once, `responseDeliverer` will drop superfluous results.
-      responseDeliverer.deliverResult(response.getResult().getAggregateFieldsMap(), readTime);
+      if (!isExplainQuery()) {
+        // Deliver the result; even though the `RunAggregationQuery` RPC is a "streaming" RPC,
+        // meaning that `onResponse()` can be called multiple times, it _should_ only be called
+        // once for non-explain queries. But even if it is called more than once,
+        // `responseDeliverer` will drop superfluous results. For explain queries, there will
+        // be more than one response, and the last response will contain the metrics.
+        onComplete();
+      }
```
tom-andersen, line 254: Can we just omit this and let the normal `onComplete` shine through? This deviation from convention doesn't add value as far as I can tell.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -75,58 +76,134 @@ public ApiFuture<AggregateQuerySnapshot> get() {
...
-    private final AtomicBoolean isFutureCompleted = new AtomicBoolean(false);
+    @Nullable private final ExplainOptions explainOptions;
+    private final SettableApiFuture<AggregateQuerySnapshot> aggregateQuerySnapshotFuture =
+        SettableApiFuture.create();
+    private final SettableApiFuture<ExplainResults<AggregateQuerySnapshot>> explainFuture =
+        SettableApiFuture.create();
+    private final AtomicBoolean isAggregateQuerySnapshotFutureCompleted = new AtomicBoolean(false);
+    private final AtomicBoolean isExplainFutureCompleted = new AtomicBoolean(false);
 
     AggregateQueryResponseDeliverer(
```
tom-andersen, line 146: There is a clear line dividing all paths of logic in this class: `if (explainOptions != null)`.

Should we maybe just have separate classes for these separate purposes. I am not sure whether this should be done with inheritance, interfaces, abstract classes, restructuring logic to AggregateQueryResponseObserver.

Current implementation seems awkward though, with some fields only be used in one path, and other fields only being used in other path.

tom-andersen (changes requested):
There are a few points where the observer pattern implementation breaks down in my opinion. Otherwise, looks good!

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
ehsannas: I agree.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -112,52 +143,156 @@ private final class AggregateQueryResponseDeliverer {
...
-        data.forEach((serverAlias, value) -> mappedData.put(aliasMap.get(serverAlias), value));
-        future.set(new AggregateQuerySnapshot(AggregateQuery.this, readTime, mappedData));
-      }
+    void setFuture(T value) {
+      future.set(value);
     }
 
     void deliverError(Throwable throwable) {
-      if (isFutureCompleted.compareAndSet(false, true)) {
+      if (futureNotCompleted()) {
```
tom-andersen, line 178: The `setException` has built in control to make sure it is only called once. From docs:

> When a call to this method returns, the Future is guaranteed to be done only if the call was accepted (in which case it returns true).

The `AtomicBoolean` is not required.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -112,52 +143,156 @@ private final class AggregateQueryResponseDeliverer {
...
+    @Override
+    void deliverResult(
+        @Nullable Map<String, Value> serverData,
+        Timestamp readTime,
+        @Nullable ExplainMetrics metrics) {
+      if (futureNotCompleted()) {
+        if (serverData == null) {
+          throw new RuntimeException("Did not receive any aggregate query results.");
+        }
+        setFuture(
```
tom-andersen, line 206: The `setFuture` has built in control to make sure it is only called once. From docs:

> When a call to this method returns, the Future is guaranteed to be done only if the call was accepted (in which case it returns true).

The `AtomicBoolean` is not required.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -112,52 +143,156 @@ private final class AggregateQueryResponseDeliverer {
...
       return future;
     }
 
-    void deliverResult(@Nonnull Map<String, Value> data, Timestamp readTime) {
-      if (isFutureCompleted.compareAndSet(false, true)) {
-        Map<String, Value> mappedData = new HashMap<>();
-        data.forEach((serverAlias, value) -> mappedData.put(aliasMap.get(serverAlias), value));
-        future.set(new AggregateQuerySnapshot(AggregateQuery.this, readTime, mappedData));
-      }
+    void setFuture(T value) {
```
tom-andersen, line 173: This should be `protected` since it is only meant for usage in child classes.

tom-andersen (changes requested):
I will need some more time, but here is a preliminary review.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
ehsannas: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
ehsannas: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
ehsannas: Done

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
tom-andersen: I agree with point about optimization.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/ExplainMetrics.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.firestore;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/** ExplainMetrics contains information about planning and execution of a query. */
+public class ExplainMetrics {
```
tom-andersen, line 22: Could we make this a `final` class?

tom-andersen (approved):
Looks good. Consider anything remaining as `nits`.

Beyond this PR, this library needs better logging. I would like to see that every usage of `SettableApiFuture` also include conditional logging if `set` or `setException` returns false. These represent unexpected situations that we are simply ignoring right now.

PR #1637, opened 2024-04-03
fix: Minor improvement to the ITE2ETracingTest.

ehsannas:


google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITE2ETracingTest.java
```diff
@@ -1193,7 +1192,6 @@ public void transactionRollbackTraceTest() throws Exception {
 
     fetchAndValidateTrace(
         customSpanContext.getTraceId(),
-        /*numExpectedSpans=*/ 5,
```
jimit-j-shah, line 1196: How does the method accurately verify number of distinct span in a transactional op trace? Cannot infer the counts from the list of compareTos - Ex. RUN is the same parent for all call stacks.

ehsannas: It [does so](https://github.com/googleapis/java-firestore/pull/1637/files#diff-a5b5f73829ce6ed82032a3839d9dabc93e911860141ac042306ef5ccfced63cbR452) by finding the number of _unique_ spans that are passed in. For example, in this case we have 6 spans:

                SPAN_NAME_TRANSACTION_RUN,
                SPAN_NAME_TRANSACTION_BEGIN,
                grpcSpanName(BEGIN_TRANSACTION_RPC_NAME)
                SPAN_NAME_TRANSACTION_RUN, <-- this is duplicate, so it's not counted
                SPAN_NAME_TRANSACTION_ROLLBACK,
                grpcSpanName(ROLLBACK_RPC_NAME)

and once we remove the duplicates we get 5.

jimit-j-shah: I think this check is broken in that if there were two callstacks with same methods but different arguments called in a transaction, uniqueness would reduce the number of expected total spans. Which is why I decoupled it from the input spans.

jimit-j-shah (approved):

PR #1639, opened 2024-04-03
fix: Rename 'enabled' to 'tracingEnabled'.

ehsannas:


google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOpenTelemetryOptions.java
```diff
@@ -75,7 +75,7 @@ public FirestoreOpenTelemetryOptions build() {
      */
     @Nonnull
     public FirestoreOpenTelemetryOptions.Builder setTracingEnabled(boolean enable) {
```
kolea2, line 77: nit - maybe rename the param too for consistency

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOpenTelemetryOptions.java
```diff
@@ -21,16 +21,16 @@
...
 
   FirestoreOpenTelemetryOptions(Builder builder) {
-    this.enabled = builder.enabled;
+    this.tracingEnabled = builder.tracingEnabled;
     this.openTelemetry = builder.openTelemetry;
   }
 
-  public boolean getEnabled() {
-    return enabled;
+  public boolean getTracingEnabled() {
```
kolea2, line 32: while we're here - I would rename this to `isTracingEnabled`

ehsannas: Good point. Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOpenTelemetryOptions.java
ehsannas: SG! Done.

kolea2 (approved):

PR #1695, opened 2024-05-24
chore: Update codebase formatting tool and introduce metalava

wu-hui:
This PR includes:
1. A new `tools` module that is only used to run metalava and generate api.txt.
2. Introduce spotless plugin to run auto format for potential kotlin files.


tom-andersen (approved):
LGTM

PR #1223, opened 2023-03-04
Add support for java record serialization

eranl:
With this change, java records will be serialized using their automatic constructors and accessors. Since records are not supported in JDK8, reflection is used for inspecting record metadata.

I created a new google-cloud-firestore-jdk17 module, for two purposes:
1. Test classes for records, which require jdk17
2. DocumentId, PropertyName, and ServerTimestamp annotations with @Target(ElementType.RECORD_COMPONENT), which also require jdk17

Fixes #1152 ☕️

Some comments and questions:
1. Since only automatic constructors are supported, exclusion of fields is not supported
2. Custom getters are not supported
3. The existing ThrowOnExtraProperties & IgnoreExtraProperties annotations are supported
4. I put the new logic outside of CustomClassMapper for readability, and extracted some of the existing logic out of it, for reuse. Everything is package-private. It can all be easily moved (back) into CustomClassMapper.
5. I based the tests on existing tests for POJOs, and adapted them to records
6. Two of the test classes use the com.google.cloud.firestore java package, in order to have access to FirestoreImpl, which is needed for mocking
7. Should the new google-cloud-firestore-jdk17 module be moved to a separate project?
8. The code currently accepts any annotations with the simple names DocumentId, PropertyName, and ServerTimestamp, for flexibility. That can be changed to require the FQNs.

PR #1706, opened 2024-06-13
Feat: Record serialization/deserialization

milaGGL:
Fixes https://github.com/googleapis/java-firestore/issues/1152
Contribution by @eranl: https://github.com/googleapis/java-firestore/pull/1223
Internal link for details: b/363272745


tom-andersen (approved):
Looks good.

My comments are mostly about internal location of code. I feel code is being split across many files, and this has caused some duplication, decrease in encapsulation and code that jumps between files more. I think this could be simplified.

If this can be cleaned up a little now, that would be good. But I don't consider this a blocker to merging. We can followup on this in future PRs if you like.

eranl:
> Looks good.
> 
> My comments are mostly about internal location of code. I feel code is being split across many files, and this has caused some duplication, decrease in encapsulation and code that jumps between files more. I think this could be simplified.
> 
> If this can be cleaned up a little now, that would be good. But I don't consider this a blocker to merging. We can followup on this in future PRs if you like.

@tom-andersen, @milaGGL, let me know if you'd like me to take a stab at these comments, either pre- or post-merge.

milaGGL:
> > Looks good.
> > My comments are mostly about internal location of code. I feel code is being split across many files, and this has caused some duplication, decrease in encapsulation and code that jumps between files more. I think this could be simplified.
> > If this can be cleaned up a little now, that would be good. But I don't consider this a blocker to merging. We can followup on this in future PRs if you like.
> 
> @tom-andersen, @milaGGL, let me know if you'd like me to take a stab at these comments, either pre- or post-merge.

hi @eranl, I agree with Tom's comments and prefer to address them pre-merge. I will work on the comments tomorrow, but you are very much welcome to help out if you would like to.

google-cloud-firestore/src/test-jdk17/java/com/google/cloud/firestore/RecordDocumentReferenceTest.java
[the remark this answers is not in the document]
milaGGL: As the main purpose of this file is testing record types using DocumentReference, having it highlighted in the file name makes more sense. in the future, we might add more tests that can only be run on jdk17+ env , and they should also specify what they are testing.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
```diff
@@ -630,32 +630,32 @@ private static <T> T convertBean(Object o, Class<T> clazz, DeserializeContext co
...
 
+  private static boolean isRecordType(Class<?> cls) {
+    Class<?> parent = cls.getSuperclass();
+    return parent != null && "java.lang.Record".equals(parent.getName());
+  }
+
   // Helper class to convert from maps to custom objects (Beans), and vice versa.
-  private static class BeanMapper<T> {
-    private final Class<T> clazz;
+  private static class PojoBeanMapper<T> extends BeanMapper<T> {
```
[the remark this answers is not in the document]
milaGGL: - [ ] moved PojoBeanMapper to separate class
- [ ] move serialization related files into `serialization` folder
- [ ] "Many methods have had their private attribute removed" 
- [ ] move duplictated code to `BeanMapper`
- [ ] move `serializeError ` and `deserializeError ` to `DeserializeContext`

google-cloud-firestore/src/main/java/com/google/cloud/firestore/encoding/CustomClassMapper.java
```diff
@@ -38,7 +43,7 @@
 import java.util.concurrent.ConcurrentMap;
 
 /** Helper class to convert to/from custom POJO classes and plain Java types. */
-class CustomClassMapper {
+public class CustomClassMapper {
```
eranl, line 46: Is it a good idea to make this class and the methods below public, exposing them to library users?

milaGGL: I am also hesitating over this.

milaGGL: added "@InternalApi" to avoid public use.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/encoding/BeanMapper.java
```diff
@@ -93,18 +90,6 @@ abstract T deserialize(
       Map<TypeVariable<Class<T>>, Type> types,
       DeserializeContext context);
 
-  void ensureValidDocumentIdType(String fieldDescription, String operation, Type type) {
```
eranl: Why move these methods from base class to a utility class?

milaGGL: I am moving out all util methods that are not specific to the base class or used in both child classes.

milaGGL: and this is renamed to `validateDocumentIdType` to match `validateServerTimestampType` method.

milaGGL:
Have done some code re-format. Please review @tom-andersen, @eranl.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CustomClassMapper.java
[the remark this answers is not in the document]
eranl: > The BeanMapper class could be a better place for common logic, since methods can be `protected` here.

I agree. Having mapper subclasses call an external utility class is less modular.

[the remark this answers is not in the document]
milaGGL: make sense

google-cloud-firestore/src/main/java/com/google/cloud/firestore/encoding/RecordMapper.java
```diff
@@ -0,0 +1,237 @@
...
+      Map<String, Object> values,
+      Map<TypeVariable<Class<T>>, Type> types,
+      DeserializeContext context) {
+    if (constructor == null) {
+      throw context.errorPath.deserializeError(
+          "Class "
+              + getClazz().getName()
+              + " does not define a no-argument constructor. If you are using ProGuard, make "
+              + "sure these constructors are not stripped");
+    }
```
eranl, line 110: This check is redundant, since this class can't be instantiated with a null `constructor`. Also, it's not a no-argument constructor, but the record's canonical one.

google-cloud-firestore/src/test-jdk17/java/com/google/cloud/firestore/RecordDocumentReferenceTest.java
```diff
@@ -0,0 +1,407 @@
...
+  @Test
+  public void serializeBasicTypes() throws Exception {
+    doReturn(SINGLE_WRITE_COMMIT_RESPONSE)
+        .when(firestoreMock)
+        .sendRequest(
+            commitCapture.capture(), ArgumentMatchers.<UnaryCallable<CommitRequest, CommitResponse>>any());
+
+    documentReference.set(ALL_SUPPORTED_TYPES_OBJECT).get();
+
+    CommitRequest expectedCommit = commit(set(ALL_SUPPORTED_TYPES_PROTO));
```
eranl, line 91: Since these files are only compile-able with JDK16+ anyway, why not use `var`?

milaGGL: Explicitly writing out the types is more readable and maintainable. Clarity is preferred over simplicity.

eranl:
Looks good. Thanks for pushing this through.

PR #1728, opened 2024-07-01
feat: Tracing using OpenTelemetry

ehsannas:


google-cloud-firestore/pom.xml
```diff
@@ -16,6 +16,7 @@
   </parent>
   <properties>
     <site.installationModule>google-cloud-firestore</site.installationModule>
+    <opentelemetry.version>1.38.0</opentelemetry.version>
```
jimit-j-shah, line 19: How does this work with the overall BOM version? Is this the minimum version required or does this need to match the BOM version?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CollectionReference.java
```diff
@@ -129,53 +128,59 @@ public DocumentReference document(@Nonnull String childPath) {
...
+              };
+            }
+          };
+      span.end();
+      return result;
     } catch (ApiException exception) {
-      span.setStatus(Status.UNKNOWN.withDescription(exception.getMessage()));
+      span.end(exception);
       throw FirestoreException.forApiException(exception);
-    } finally {
```
jimit-j-shah, line 151: Why remove the `finally`? (Though the new implementation looks more correct since it doesn't call `end` multiple times when no exception)

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOpenTelemetryOptions.java
```diff
@@ -0,0 +1,96 @@
...
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore;
+
+import io.opentelemetry.api.OpenTelemetry;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class FirestoreOpenTelemetryOptions {
```
jimit-j-shah, line 23: Block comment briefly describing the class?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/DisabledTraceUtil.java
```diff
@@ -0,0 +1,108 @@
...
+package com.google.cloud.firestore.telemetry;
+
+import com.google.api.core.ApiFunction;
+import com.google.api.core.ApiFuture;
+import io.grpc.ManagedChannelBuilder;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class DisabledTraceUtil implements TraceUtil {
```
jimit-j-shah, line 26: Block comment with brief description of the class?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledTraceUtil.java
```diff
@@ -0,0 +1,390 @@
...
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.instrumentation.grpc.v1_6.GrpcTelemetry;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class EnabledTraceUtil implements TraceUtil {
```
jimit-j-shah, line 39: Block comment with brief description of the class?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/TraceUtil.java
```diff
@@ -0,0 +1,162 @@
...
+
+import com.google.api.core.ApiFunction;
+import com.google.api.core.ApiFuture;
+import com.google.cloud.firestore.FirestoreOptions;
+import io.grpc.ManagedChannelBuilder;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public interface TraceUtil {
```
jimit-j-shah, line 27: Block comment with brief description of the interface?

jimit-j-shah (approved):

google-cloud-firestore/pom.xml
ehsannas: below we use

```
    <dependency>
      <groupId>io.opentelemetry</groupId>
      <artifactId>opentelemetry-api</artifactId>
    </dependency>
    <dependency>
      <groupId>io.opentelemetry</groupId>
      <artifactId>opentelemetry-context</artifactId>
    </dependency>
    <dependency>
      <groupId>io.opentelemetry.instrumentation</groupId>
      <artifactId>opentelemetry-grpc-1.6</artifactId>
    </dependency>
```

without specifying any version. The version for these libraries will be picked up from the BOM.

This (`<opentelemetry.version>1.38.0</opentelemetry.version>`) is only used below by the libraries that are used for testing (the ones that have `<scope>test</scope>`). Not all of these libraries are in the "shared dependencies". So, if there is a major version change in the otel version of the BOM, and we don't update the version used by our tests, then our tests will fail (or not compile) and we'll need to update this version before being able to release.

I'll ask about whether there's a better approach for shared package versioning for testing (maybe we should move all the needed testing libraries to the shared bom as well?). I think this is acceptable for this PR, and I have filed b/353596681 to follow up.

PR #1826, opened 2024-09-17
fix: Remove race condition in test.

tom-andersen:
CI sometimes fails on `listenToDocumentsWithVectors`, and as best as I can tell, it is because of race condition in test.

There is a blocking call on listener that precedes incrementing semaphore that controls switch statement. Consequently, the case that deletes the document runs twice. The second time, the document has been deleted, so assertion fails on `assertNotNull(docSnap)` line 1456. This can happen if the executor on listener is a thread pool capable of concurrent execution, allowing next snapshot to process, before previous snapshot is done processing.

This PR makes the code more readable, and removes any chance of race condition.

I have applied this to both places where this pattern appears.


MarkDuckworth (approved):
LGTM. Thanks for the fix!

PR #1827, opened 2024-09-17
feat: Add support for vector search with Query#findNearest

MarkDuckworth:

Fixes #<issue_number_goes_here> ☕️

If you write sample code, please follow the [samples format](
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md).


google-cloud-firestore/src/main/java/com/google/cloud/firestore/FindNearestOptions.java
```diff
@@ -0,0 +1,83 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents the options that are used to configure the use of OpenTelemetry for telemetry
```
wu-hui, line 22: Copy/Paste error?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/GenericQuerySnapshot.java
```diff
@@ -0,0 +1,161 @@
...
+  volatile List<QueryDocumentSnapshot> documents;
+
+  volatile DocumentSet documentSet;
+
+  protected GenericQuerySnapshot(
+      QueryT query,
+      Timestamp readTime,
+      final List<QueryDocumentSnapshot> documents,
+      final DocumentSet documentSet,
+      final List<DocumentChange> documentChanges) { // Elevated access level for mocking.
```
wu-hui, line 39: Nit: the comment should be on top of the constructor.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/QueryApiStreamObserver.java
```diff
@@ -0,0 +1,49 @@
...
+package com.google.cloud.firestore;
+
+import com.google.api.gax.rpc.ApiStreamObserver;
+import com.google.cloud.Timestamp;
+import com.google.firestore.v1.Document;
+import com.google.firestore.v1.RunQueryResponse;
+import java.util.ArrayList;
+import java.util.List;
+
+public abstract class QueryApiStreamObserver implements ApiStreamObserver<RunQueryResponse> {
```
wu-hui, line 26: Add comments, and annotate to tell if this is part of the public API or for internal use only?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
```diff
@@ -0,0 +1,400 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.threeten.bp.Duration;
+
+/**
+ * Represents a query whose results can be streamed. If the stream fails with a retryable error,
+ * implementations of IStreamableQuery can optionally support retries with a cursor, as indicated by
```
wu-hui, line 48: IStreamableQuery -> StreamableQuery.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -2128,6 +1824,60 @@ public AggregateQuery aggregate(
     return new AggregateQuery(this, aggregateFieldList);
   }
 
+  public VectorQuery findNearest(
```
wu-hui, line 1827: Don't we need some public javadoc for these? Same with VectorQuery.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -332,8 +317,8 @@ public boolean equals(Object o) {
   }
 
   static final class FieldOrder implements Comparator<QueryDocumentSnapshot> {
-    private final FieldReference fieldReference;
-    private final Direction direction;
+    final FieldReference fieldReference;
+    final Direction direction;
```
tom-andersen, line 322: Why are these no longer private?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -2128,6 +1823,173 @@ public AggregateQuery aggregate(
...
+      VectorValue queryVector,
+      int limit,
+      VectorQuery.DistanceMeasure distanceMeasure,
+      VectorQueryOptions vectorQueryOptions) {
+    if (limit <= 0) {
+      throw FirestoreException.forInvalidArgument(
+          "Not a valid positive `limit` number. `limit` must be larger than 0.");
+    }
+
+    if (queryVector.toArray().length == 0) {
```
tom-andersen, line 1984: Could we put a size method on `VectorValue`? The `toArray()` method clones the array, which seems excessive when all we want to know is the length.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -2128,6 +1823,173 @@ public AggregateQuery aggregate(
...
+   * @return an {@link VectorQuery} that performs vector distance (similarity) search with the given
+   *     parameters.
+   */
+  public VectorQuery findNearest(
+      FieldPath vectorField,
+      VectorValue queryVector,
+      int limit,
+      VectorQuery.DistanceMeasure distanceMeasure) {
+    return findNearest(
+        vectorField, queryVector, limit, distanceMeasure, VectorQueryOptions.newBuilder().build());
```
tom-andersen, line 1941: This seems frequently calculated: `VectorQueryOptions.newBuilder().build()`

Should we create constant `VectorQueryOptions.DEFAULT` ?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/QuerySnapshot.java
```diff
@@ -104,134 +49,16 @@ public static QuerySnapshot withChanges(
...
-    for (DocumentSnapshot documentSnapshot : documents) {
-      results.add(
-          CustomClassMapper.convertToCustomClass(
-              documentSnapshot.getData(), clazz, documentSnapshot.getReference()));
-    }
-
-    return results;
+  public Query getQuery() {
+    return super.getQuery();
   }
```
tom-andersen, line 63: This `Override` doesn't seem to do anything and could just be removed. It only serves to remove `@Nonnull` annotation from super. Is that intentional?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuerySnapshot.java
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Returns the VectorQuery for the snapshot.
+   *
+   * @return The backing query that produced this snapshot.
+   */
+  @Override
+  public VectorQuery getQuery() {
+    return super.getQuery();
+  }
```
tom-andersen, line 53: This Override doesn't seem to do anything and could just be removed. It only serves to remove @Nonnull annotation from super. Is that intentional?

```suggestion
```

tom-andersen (changes requested):
I am still going through PR, but I want to share what I found so far.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1232,6 +1211,11 @@ public Query select(FieldPath... fieldPaths) {
     return new Query(rpcContext, newOptions.build());
   }
 
+  @Override
+  public boolean isRetryableWithCursor() {
```
tom-andersen, line 1215: I don't think we want this as part of our API surface.

```suggestion
  boolean isRetryableWithCursor() {
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuery.java
```diff
@@ -0,0 +1,191 @@
...
+            StructuredQuery.FindNearest.DistanceMeasure.EUCLIDEAN);
+        break;
+      case DOT_PRODUCT:
+        findNearestBuilder.setDistanceMeasure(
+            StructuredQuery.FindNearest.DistanceMeasure.DOT_PRODUCT);
+        break;
+      default:
+        findNearestBuilder.setDistanceMeasure(
+            StructuredQuery.FindNearest.DistanceMeasure.UNRECOGNIZED);
+    }
```
tom-andersen, line 140: I think a helper method could cleanup this code, and remove some repetition:

```
  private static StructuredQuery.FindNearest.DistanceMeasure toProto(DistanceMeasure distanceMeasure) {
    switch (distanceMeasure) {
      case COSINE:
        return StructuredQuery.FindNearest.DistanceMeasure.COSINE;
      case EUCLIDEAN:
        return StructuredQuery.FindNearest.DistanceMeasure.EUCLIDEAN;
      case DOT_PRODUCT:
        return StructuredQuery.FindNearest.DistanceMeasure.DOT_PRODUCT;
      default:
        return StructuredQuery.FindNearest.DistanceMeasure.UNRECOGNIZED;
    }
  }
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/GenericQuerySnapshot.java
```diff
@@ -0,0 +1,173 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore;
+
+import com.google.cloud.Timestamp;
+import com.google.cloud.firestore.encoding.CustomClassMapper;
+import com.google.common.collect.ImmutableList;
+import java.util.*;
```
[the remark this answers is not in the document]
MarkDuckworth: Thanks. Fixed. My IDE seems to be opinionated about this, but with the opposite opinion.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/GenericQuerySnapshot.java
```diff
@@ -0,0 +1,173 @@
...
+              ImmutableList.builderWithExpectedSize(size);
+          for (int i = 0; i < size; ++i) {
+            builder.add(new DocumentChange(documents.get(i), DocumentChange.Type.ADDED, -1, i));
+          }
+          documentChanges = builder.build();
+        }
+      }
+    }
+
+    return Collections.unmodifiableList(documentChanges);
```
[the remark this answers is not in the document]
MarkDuckworth: Good call. No longer needed with the constructor update.

tom-andersen (changes requested):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/GenericQuerySnapshot.java
```diff
@@ -0,0 +1,173 @@
...
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    GenericQuerySnapshot that = (GenericQuerySnapshot) o;
+    return Objects.equals(query, that.query)
+        && Objects.equals(this.size(), that.size())
```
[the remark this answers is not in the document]
MarkDuckworth: good call. Removed.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/GenericQuerySnapshot.java
```diff
@@ -0,0 +1,173 @@
...
+    return getDocuments().size();
+  }
+
+  /**
+   * Tests for equality with this object.
+   *
+   * @param o is tested for equality with this object.
+   * @return `true` if equal, otherwise `false`
+   */
+  public boolean equals(Object o) {
```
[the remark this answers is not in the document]
MarkDuckworth: I'm not sure. This was existing logic in the QuerySnapshot that is abstracted out to a base GenericQuerySnapshot.

I suspect though, that adding readTime to `equals(...)` effectively makes testing value equality pointless. There would never be two instances of a QuerySnapshot that are equal unless they are the same object.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/GenericQuerySnapshot.java
```diff
@@ -0,0 +1,173 @@
...
+  volatile List<QueryDocumentSnapshot> documents;
+
+  volatile DocumentSet documentSet;
+
+  // Elevated access level for mocking.
+  protected GenericQuerySnapshot(
+      QueryT query,
+      Timestamp readTime,
+      final List<QueryDocumentSnapshot> documents,
+      final DocumentSet documentSet,
```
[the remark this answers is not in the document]
MarkDuckworth: Also pre-existing logic, so I'm not certain of the reasoning, but there are several instances of lazy-loading and memoization in this code. My suspicion is that it's related to performance. I don't have strong feelings on this either way. What are your thoughts with that context?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
MarkDuckworth: Artifact of early approach to refactoring. Thanks!

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
MarkDuckworth: Adding a package private `size()` method for efficency and consistency with naming in the SDK. Will get API review before making this or something similar public.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
MarkDuckworth: I added a package private `getDefaultInstance()`. That's consistent naming with `FirestoreOptions.getDefaultInstance()`. I can make this public with API review, but I'm not sure what value it adds. We have overloads so customers don't need a default instance.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/QuerySnapshot.java
MarkDuckworth: Without this, the clirr style check fails reporting ` Method 'public com.google.cloud.firestore.Query getQuery()' has been removed`. It's effectively the same, but `getQuery()` on the base is generic.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuerySnapshot.java
MarkDuckworth: Same, although I'm adding @Nonnull to both

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
MarkDuckworth: Great catch. Thanks.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuery.java
MarkDuckworth: Yes! Very nice. Thanks.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuerySnapshot.java
MarkDuckworth: Following up on our discussion offline, I removed both of these overrides.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/QuerySnapshot.java
MarkDuckworth: Following up on our discussion offline, I removed both of these overrides.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/GenericQuerySnapshot.java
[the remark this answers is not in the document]
tom-andersen: Since this is existing logic, I am happy to leave alone.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQueryOptions.java
```diff
@@ -0,0 +1,153 @@
...
+        && Objects.equals(distanceThreshold, otherOptions.distanceThreshold);
+  }
+
+  /**
+   * Returns a default {@code FirestoreOptions} instance. Note: package private until API review can
+   * be completed.
+   */
+  static VectorQueryOptions getDefaultInstance() {
+    return newBuilder().build();
+  }
```
tom-andersen, line 152: This is to avoid instantiating the same object over and over again.

```suggestion
  private static DEFAULT = newBuilder().build();

  /**
   * Returns a default {@code FirestoreOptions} instance. Note: package private until API review can
   * be completed.
   */
  static VectorQueryOptions getDefaultInstance() {
    return DEFAULT;
  }
```

google-cloud-firestore/src/test/java/com/google/cloud/firestore/VectorQueryTest.java
```diff
@@ -0,0 +1,474 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.firestore;
+
+import static com.google.cloud.firestore.LocalFirestoreHelper.*;
+import static org.junit.Assert.*;
```
tom-andersen, line 20: nit: This is another wildcard import.

Since this is in a test file, I am not as concerned.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1705,148 +1710,12 @@ private Value encodeValue(FieldPath fieldPath, Object value) {
...
-
-    rpcContext.streamRequest(request.build(), observer, rpcContext.getClient().runQueryCallable());
-  }
-
   /**
    * Executes the query and returns the results as QuerySnapshot.
    *
    * @return An ApiFuture that will be resolved with the results of the Query.
    */
-  @Nonnull
```
tom-andersen, line 1849: I think we still want this to be `@Nonnull` ?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
```diff
@@ -1859,79 +1728,9 @@ public ApiFuture<QuerySnapshot> get() {
    * @return An ApiFuture that will be resolved with the planner information, statistics from the
    *     query execution (if any), and the query results (if any).
    */
-  @Nonnull
```
tom-andersen, line 1862: Same here.

I think we still want this to be `@Nonnull` ?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuery.java
```diff
@@ -0,0 +1,190 @@
...
+    this.distanceMeasure = distanceMeasure;
+  }
+
+  /**
+   * Executes the query and returns the results as {@link QuerySnapshot}.
+   *
+   * @return An ApiFuture that will be resolved with the results of the VectorQuery.
+   */
+  @Override
+  public ApiFuture<VectorQuerySnapshot> get() {
```
tom-andersen, line 65: I think we still want this to be `@Nonnull` ?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuery.java
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Plans and optionally executes this VectorQuery. Returns an ApiFuture that will be resolved with
+   * the planner information, statistics from the query execution (if any), and the query results
+   * (if any).
+   *
+   * @return An ApiFuture that will be resolved with the planner information, statistics from the
+   *     query execution (if any), and the query results (if any).
+   */
+  @Override
+  public ApiFuture<ExplainResults<VectorQuerySnapshot>> explain(ExplainOptions options) {
```
tom-andersen, line 78: Same. I think we still want this to be `@Nonnull` ?

tom-andersen (changes requested):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
```diff
@@ -0,0 +1,400 @@
...
+  }
+
+  abstract RunQueryRequest.Builder toRunQueryRequestBuilder(
+      @Nullable final ByteString transactionId,
+      @Nullable final Timestamp readTime,
+      @Nullable ExplainOptions explainOptions);
+
+  abstract boolean isRetryableWithCursor();
+
+  abstract StreamableQuery startAfter(@Nonnull DocumentSnapshot snapshot);
```
tom-andersen, line 68: I realize this is pre-existing, but it really should be fixed.

Also, the `Query` class should have it's `startAfter` method marked with `@Override`

```suggestion
  abstract StreamableQuery<QuerySnapshot> startAfter(@Nonnull DocumentSnapshot snapshot);
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuery.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+  }
+
+  @Override
+  boolean isRetryableWithCursor() {
+    return false;
+  }
+
+  @Override
+  StreamableQuery startAfter(@Nonnull DocumentSnapshot snapshot) {
```
tom-andersen, line 162:
```suggestion
  VectorQuery startAfter(@Nonnull DocumentSnapshot snapshot) {
```

tom-andersen (changes requested):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQueryOptions.java
MarkDuckworth: Updated.

google-cloud-firestore/src/test/java/com/google/cloud/firestore/VectorQueryTest.java
MarkDuckworth: updated

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
MarkDuckworth: good catch. Super annoying that you can change this attribute in an override

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Query.java
MarkDuckworth: Same. Thanks.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuery.java
MarkDuckworth: same

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuery.java
MarkDuckworth: same

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
MarkDuckworth: Updated. Thanks.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/VectorQuery.java
MarkDuckworth: Updated

google-cloud-firestore/src/main/java/com/google/cloud/firestore/GenericQuerySnapshot.java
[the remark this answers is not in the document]
MarkDuckworth: I changed this.

I don't think there was an issue before. The value of `documentSet` was tested for equality because it's value was converted by the `getDocuments()` method.

However, I think the new approach is equally fine. The benefits of lazily converting `documentSet` to a list of document snapshots are rarely if ever seen.

tom-andersen (approved):
Nice!

PR #1840, opened 2024-09-19
Feat: client side metrics

milaGGL:
- [ ] Integrate Opentelemetry API and SDK into the Firestore SDK
- [ ] Implement `BuiltinMetricsProvider` class to collect 4 GAX layer and 4 SDK layer metrics
- [ ] Integrate `GoogleCloudMetricExporter` to export metrics data to CloudMonitoring `custom.googleapis.com` namespace (Waiting for BE setup)
- [ ] Create public API to enable/disable client side metrics (the API is set to private, and disaled for now.)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/BuiltinMetricsConstants.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class BuiltinMetricsConstants {
+
+  static final String FIRESTORE_LIBRARY_NAME = "firestore_java";
+
+  // TODO: change to firestore.googleapis.com
+  public static final String METER_NAME = "custom.googleapis.com/internal/client";
```
milaGGL, line 34: Opentelemetry uses dpt seperation in metrics naming, while monarch uses slash separation. exporting the built-in metrics to other backends might raise naming conflict.

milaGGL: we could have named the metrics with dot seperation, use view to reformat it with slash separation in the default OTEL instance for monarch. But the metrics from Gax uses slash already, using dot separation inside firestore doesn't solves the problem.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/MetricsUtil.java
```diff
@@ -0,0 +1,86 @@
...
+     * of the given future.
+     */
+    <T> void recordEndToEndLatencyAtFuture(ApiFuture<T> futureValue);
+
+    void recordEndToEndLatency();
+
+    void recordEndToEndLatency(Throwable t);
+
+    /** Records first response latency for the current operation. */
+    void recordFirstResponseLatency();
```
milaGGL, line 84: To be able to record first response latency, we should go inside the "internalStream" function, which is called by other public APIs, ie: Query.stream, explan, get, or recursively inside internalStream on error.  The mthod context will be lost with the current design, need to come up with a better way to pass the method info and stop watch down into the call, like tracing do.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -427,16 +431,30 @@ public <T> ApiFuture<T> runAsyncTransaction(
...
-      // READ_ONLY transactions with readTime have no retry, nor transaction state, so we don't need
-      // a runner.
-      return updateFunction.updateCallback(
-          new ReadTimeTransaction(this, transactionOptions.getReadTime()));
-    } else {
-      // For READ_ONLY transactions without readTime, there is still strong consistency applied,
-      // that cannot be tracked client side.
-      return new ServerSideTransactionRunner<>(this, updateFunction, transactionOptions).run();
+    MetricsContext metricsContext =
+        getOptions().getMetricsUtil().createMetricsContext("RunTransaction");
```
milaGGL, line 435: some of the function names are not in SPAN_NAME.. list. add them on.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -249,6 +250,9 @@ public void onResponse(BatchGetDocumentsResponse response) {
             DocumentReference documentReference;
             DocumentSnapshot documentSnapshot;
 
+            // TODO(metrics): record first_response_latency here. Need to come up with a way to
```
milaGGL, line 253: Some methods in this calls should have added monitoring info, like `recursiveDelete`, `getAll`, `runTransaction`. discuss with Ehsan.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledMetricsUtil.java
```diff
@@ -0,0 +1,232 @@
...
+
+      OpenTelemetry customOpenTelemetry =
+          firestoreOptions.getOpenTelemetryOptions().getOpenTelemetry();
+      if (customOpenTelemetry == null) {
+        customOpenTelemetry = GlobalOpenTelemetry.get();
+      }
+      this.customMetricsProvider = new BuiltinMetricsProvider(customOpenTelemetry);
+    } catch (IOException e) {
+      logger.warning(
+          "Unable to get create MetricsUtil object for client side metrics, will skip exporting client side metrics"
```
ehsannas, line 86: nit: `Unable to create MetricsUtil ...`

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledMetricsUtil.java
```diff
@@ -0,0 +1,232 @@
...
+      attributes.put(METRIC_KEY_METHOD.getKey(), methodName);
+      attributes.put(METRIC_KEY_STATUS.getKey(), status);
+      return attributes;
+    }
+
+    private String extractErrorStatus(@Nullable Throwable throwable) {
+      if (!(throwable instanceof FirestoreException)) {
+        return StatusCode.Code.UNKNOWN.toString();
+      }
+      Status status = ((FirestoreException) throwable).getStatus();
```
ehsannas, line 228: `FirestoreException::getStatus()` may return `null`. So the code should check for that

```java
Status status = ((FirestoreException) throwable).getStatus();
if (status == null) {
  return StatusCode.Code.UNKNOWN.toString();
}
return status.getCode().name();
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledMetricsUtil.java
```diff
@@ -0,0 +1,232 @@
...
+    // MetricExporter metricExporter =
+    //     GoogleCloudMetricExporter.createWithConfiguration(
+    //         MetricConfiguration.builder()
+    //             .setProjectId(projectId)
+    //             // Ignore library info as it is collected by the metric attributes as well
+    //             .setInstrumentationLibraryLabelsEnabled(false)
+    //             .setMonitoredResourceDescription(monitoredResourceMapping)
+    //             .setUseServiceTimeSeries(true)
+    //             .build());
+
```
ehsannas, line 135: remove commented code.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/MetricsUtil.java
```diff
@@ -0,0 +1,114 @@
...
+      return new EnabledMetricsUtil(firestoreOptions);
+    } else {
+      return new DisabledMetricsUtil();
+    }
+  }
+
+  static boolean shouldCreateEnabledInstance(FirestoreOptions firestoreOptions) {
+    // Start with the value from FirestoreOptions
+    boolean shouldCreateEnabledInstance =
+        firestoreOptions.getOpenTelemetryOptions().isMetricsEnabled();
```
ehsannas, line 58: as discussed offline, we don't want `isMetricsEnabled()`. This is a large PR already, and ideally we should _disable_ the feature and land it in `main`. IIUC you can disable it by doing this here:

```java
  // TODO(metrics): Look up the enable/disable flag from options before launching
  boolean shouldCreateEnabledInstance = false;
```

ehsannas (commented):
Great work @milaGGL. A few minor comments/nits below.

I'd like us to disable the feature and land this PR

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/ClientIdentifier.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private static String getProcessId() {
+    try {
+      // Check if Java 9+ and ProcessHandle class is available
+      Class<?> processHandleClass = Class.forName("java.lang.ProcessHandle");
+      Method currentMethod = processHandleClass.getMethod("current");
+      Object processHandleInstance = currentMethod.invoke(null);
+      Method pidMethod = processHandleClass.getMethod("pid");
+      long pid = (long) pidMethod.invoke(processHandleInstance);
```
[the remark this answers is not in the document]
milaGGL: both BigTable and Spanner used `UUID+PID+Hostname` format. we could follow the same, or use sth else instead of PID.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreOpenTelemetryOptions.java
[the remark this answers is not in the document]
milaGGL: I am waiting for the API changes to be merged in: https://github.com/googleapis/java-firestore/pull/1869. 
But I can remove it now.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -339,17 +366,17 @@ public void onError(Throwable throwable) {
...
         runQuery(responseDeliverer, attempt + 1);
       } else {
         getTraceUtil()
             .currentSpan()
             .addEvent(
-                SPAN_NAME_RUN_AGGREGATION_QUERY + ": Error",
+                METHOD_NAME_RUN_AGGREGATION_QUERY + ": Error",
                 Collections.singletonMap("error.message", throwable.getMessage()));
         responseDeliverer.deliverError(throwable);
+        metricsContext.recordEndToEndLatency(throwable);
```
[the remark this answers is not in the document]
milaGGL: Maybe I might misunderstood it. A query (regular or aggregation) receives "onError" signal doesn't mean that the request is completed. 
- if the error is not retiabled, the response observer will mark it as completed (like `future.setException(throwable);`) and we can record the end-to-end latency
- if the error is retriable, we will recursively invoke another `runQuery`, which carry on the same response observer, and creates another metricContext with a new start. when the retry was successful, both `runQuery` will recieve onCompleted signal, and both metricsContext will record end-to-end latency. 

end-to-end latency should include the time spend on retries.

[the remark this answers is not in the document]
milaGGL: isn't it? when retry, the previous runQuery just stops and no longer receive any responses?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/CompositeApiTracerFactory.java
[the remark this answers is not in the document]
milaGGL: The classes are to modify the GAX layer api tracing , not specific to telemetry. But sure, it is tracing after all.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/MetricsUtil.java
```diff
@@ -0,0 +1,114 @@
...
+  /**
+   * Creates an instance of {@code MetricsUtil}. If the environment variable
+   * `FIRESTORE_ENABLE_METRICS` is set to false or off, an instance of {@link DisabledMetricsUtil}
+   * will be returned. Otherwise, an instance of {@link EnabledMetricsUtil} will be returned.
+   *
+   * @param firestoreOptions The Firestore options that configures client side metrics.
+   * @return An instance of {@code MetricsUtil}.
+   */
+  static MetricsUtil getInstance(@Nonnull FirestoreOptions firestoreOptions) {
+    // TODO(metrics): add debug level logging if possible
```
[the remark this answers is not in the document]
milaGGL: in mobile SDKs, we have debug level logging. I wonder if the server SDKs would also have such logging system.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/ClientIdentifier.java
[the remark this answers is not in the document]
ehsannas: We can keep it the same. The PID is guaranteed to be unique per vm, so UUID+PID+Hostname makes sense.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/MetricsUtil.java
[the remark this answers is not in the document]
ehsannas: Oh, I see. Currently the server SDKs have near-zero logging [which is not good].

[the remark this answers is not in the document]
milaGGL: maybe we should put it into backlog. debug level logging was quite helpful with mobile&web sdk issue tickets. For now, we can add a `logger.info()` if the client side metrics is enabled.

google-cloud-firestore/pom.xml
```diff
@@ -119,6 +119,15 @@
...
       <artifactId>opentelemetry-grpc-1.6</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-sdk</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.cloud.opentelemetry</groupId>
+      <artifactId>exporter-metrics</artifactId>
+      <version>0.31.0</version>
```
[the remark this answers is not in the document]
milaGGL: exporter-metrics is already added: https://github.com/googleapis/sdk-platform-java/pull/3078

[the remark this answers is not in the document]
ehsannas: In that case you should be able to remove the `<version>0.31.0</version>` from here, so that it picks up the version from the shared deps.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
```diff
@@ -233,25 +262,61 @@ public void onCompleted() {
...
+    @Override
+    public void onCompleted() {
+      metricsContext.recordEndToEndLatency();
+      observer.onCompleted();
+    }
+  }
+
   protected void internalStream(
-      final ApiStreamObserver<RunQueryResponse> runQueryResponseObserver,
+      final MonitoredStreamResponseObserver streamResponseObserver,
```
milaGGL, line 305: there are 4 functions calling `internalStream`, if we handle the metrics recording outside `internalStream`, then it is repeated for 4 times. 
If we pass `metricsContext` into the `internalStream` directly, then there will be too many parameters.

The metrics call are basically in sync with the `ApiStreamObserver` calls (onResponse, onError, onCompleted), we can group them into one monitored stream observer class.  @ehsannas, what do you think if we integrate the tracing into the new class as well?

google-cloud-firestore/clirr-ignored-differences.xml
```diff
@@ -314,4 +314,11 @@
     <method>com.google.cloud.firestore.Query getQuery()</method>
   </difference>
 
+    <!-- Relocate constants in TraceUtil -->
+  <difference>
```
milaGGL, line 318: waiting for https://github.com/googleapis/java-firestore/pull/1869 to be merged to see if "internalApi" will lift the clirr restriction on the contstants removal, and add internalStream parameter change on.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledMetricsUtil.java
milaGGL: turns out, at least in transactions, getStatus() always return null.

google-cloud-firestore/clirr-ignored-differences.xml
```diff
@@ -314,4 +314,19 @@
...
+    <differenceType>6011</differenceType>
+    <className>com/google/cloud/firestore/telemetry/TraceUtil</className>
+    <field>SPAN_NAME_*</field>
+  </difference>
+
+  <!-- Change the parameter type of internalStream function -->
+  <difference>
```
milaGGL, line 325: waiting for https://github.com/googleapis/java-firestore/pull/1896 to lift the clirr alert

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
ehsannas: that's a good idea 👍

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
```diff
@@ -239,12 +262,47 @@ public void onCompleted() {
       return result;
     } catch (Exception error) {
       span.end(error);
+      metricsContext.recordEndToEndLatency(error);
       throw error;
     }
   }
 
+  class MonitoredStreamResponseObserver implements ApiStreamObserver<RunQueryResponse> {
```
ehsannas, line 270: I like the idea of `MonitoredStreamResponseObserver`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
```diff
@@ -331,7 +379,7 @@ public void onError(Throwable throwable) {
 
               startAfter(cursor)
                   .internalStream(
-                      runQueryResponseObserver,
+                      streamResponseObserver,
```
ehsannas, line 382: we're re-using the `streamResponseObserver`, which is great. This is good since it'll end up with only 1 first-response-latency. We should verify (using a test) the following two scenarios:

```
doc1, doc2
retryable_error_1
retry with cursor
doc3
retryable_error_2
retry with cursor
doc4
completed
```

And we should verify that the following occurs:
```
doc1
streamResponseObserver.onNext --> records first-response latency
doc2
streamResponseObserver.onNext
retryable_error_1
doc3
streamResponseObserver.onNext
retryable_error_2
doc4
streamResponseObserver.onNext
streamResponseObserver.onCompleted
// assert that streamResponseObserver.onCompleted gets called exactly once
// assert that streamResponseObserver.onError does not get called
```

And the second scenario would be:

```
doc1, doc2
non-retryable-error
```
And we should verify that the following occurs:
```
doc1
streamResponseObserver.onNext --> records first-response latency
doc2
streamResponseObserver.onNext
non-retryable-error
streamResponseObserver.onError
// assert that streamResponseObserver.onCompleted does not get called
// assert that streamResponseObserver.onError gets called exactly once.
```

ehsannas (commented):
LGTM. Only nit is to add a test to verify the behavior  of `MonitoredStreamResponseObserver` in face of a retryable or non-retryable error.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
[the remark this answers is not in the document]
milaGGL: discussed offline.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
milaGGL: it should have only one first-response-latency, as the observer object itself has a `receivedFirstResponse` flag planted inside. Since we are passing down the same object, the flag should get flipped only once. 

i haven't looked into testing yet, but seems like we can mock above responses with `doAnswer`.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
ehsannas: As discussed offline, we can do it in a separate PR.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
ehsannas: Ok, I've filed b/376308036. We should add these tests as part of your testing PR.

google-cloud-firestore/clirr-ignored-differences.xml
ehsannas: I believe the decision for #1896 was to use `internalApi` rather than making the class package-private. So we'll need to keep this clirr change as well.

ehsannas (approved):

google-cloud-firestore/clirr-ignored-differences.xml
milaGGL: they are changing the `internalStream` to default from `protected`. So it would affect the public access of that method, which could remove the need to add clirr for that functions parameter change.

PR #1869, opened 2024-10-04
fix: Drop the setTracingEnabled flag from Options (@BetaApi change) 

ehsannas:
the deprecated `setTracingEnabled()` API is redundant and is being removed.

Firestore SDK will use the OpenTelemetry instance passed to it. In its absence, it will use the Global OpenTelemetry by default. There is no need to enable tracing via the `setTracingEnabled()` API. If the user wants to disable tracing, they can still do so by providing a no-op OpenTelemetry instance or by using `FIRESTORE_ENABLE_TRACING=OFF` env var.


google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/TraceUtil.java
```diff
@@ -75,7 +75,7 @@ public interface TraceUtil {
    * @return An instance of the TraceUtil class.
    */
   static TraceUtil getInstance(@Nonnull FirestoreOptions firestoreOptions) {
```
milaGGL, line 77: since we are updating the class, can we also annotate the util class internalApi? it is marked public to be accessible outside the package, but it is not a public API.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
```diff
@@ -340,15 +340,15 @@ public void onError(Throwable throwable) {
             .currentSpan()
             .addEvent(
                 SPAN_NAME_RUN_AGGREGATION_QUERY + ": Retryable Error",
-                Collections.singletonMap("error.message", throwable.getMessage()));
+                Collections.singletonMap("error.message", throwable.toString()));
```
milaGGL, line 343: what is the motivation for using toString instead of getMessage? should other files be changed as well?

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledTraceUtil.java
```diff
@@ -84,6 +85,11 @@ public ManagedChannelBuilder apply(ManagedChannelBuilder managedChannelBuilder)
   @Override
   @Nullable
   public ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder> getChannelConfigurator() {
+    // Note: using `==` rather than `.equals` since OpenTelemetry has only 1 static instance of
+    // `TracerProvider.noop`.
+    if (openTelemetry.getTracerProvider() == TracerProvider.noop()) {
```
milaGGL, line 90: IIRC, an OTEL instance can register multiple MetricsProvider, even though it did not work as expected when I try on a test app. 
I expect TracerProvider would be the same. https://opentelemetry.io/docs/specs/otel/trace/api/#tracerprovider
`some applications may want to or have to use multiple TracerProvider instances`

google-cloud-firestore/src/main/java/com/google/cloud/firestore/AggregateQuery.java
ehsannas: this is the implementation of `toString()`:

```java
    public String toString() {
        String s = getClass().getName();
        String message = getMessage();
        return (message != null) ? (s + ": " + message) : s;
    }
```

Using toString() is safer than getMessage because toString handles cases where message is null. 

This PR has fixed it for all cases related to tracing. I think there are a couple of other instances that we should fix too [in a separate PR].

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledTraceUtil.java
```diff
@@ -31,6 +31,7 @@
 import io.opentelemetry.api.trace.SpanKind;
 import io.opentelemetry.api.trace.StatusCode;
 import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.api.trace.TracerProvider;
```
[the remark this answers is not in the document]
ehsannas: Good catch. Done.

google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/TraceUtil.java
ehsannas: Good point. Done.

milaGGL (approved):
Please add PR description for future reference. Otherwise, LGTM

google-cloud-firestore/src/test/java/com/google/cloud/firestore/it/ITE2ETracingTest.java
```diff
@@ -319,20 +319,13 @@ public void before() throws Exception {
...
     // but because gRPC traces need to be deterministically force-flushed, firestore.shutdown()
     // must be called in @After for each test.
-    FirestoreOptions.Builder optionsBuilder;
-    if (isUsingGlobalOpenTelemetrySDK()) {
-      optionsBuilder =
-          FirestoreOptions.newBuilder()
-              .setOpenTelemetryOptions(
-                  FirestoreOpenTelemetryOptions.newBuilder().setTracingEnabled(true).build());
-    } else {
+    FirestoreOptions.Builder optionsBuilder = FirestoreOptions.newBuilder();
```
milaGGL, line 322: Question, I saw there are 4 classes like `ITTracingTestGlobalOtel extends ITTracingTest `, `ITTracingTestNonGlobalOtel extends ITTracingTest`, but they are not used anywhere.  Plus, the default behaviour is being changed to "use custom otel or fall back to Global otel", do we still need that 4 classes?

ehsannas: Regarding `they are not used anywhere`:

`ITTracingTest` and `ITE2ETracingTest` are abstract base classes, and therefore they actually won't run any tests. When the `it` test suite is run, it will run the inherited classes (`ITE2ETracingTestGlobalOtel`, `ITE2ETracingTestNonGlobalOtel`, `ITTracingTestGlobalOtel`, and `ITTracingTestNonGlobalOtel`). So there are 2 test suites, each one runs in two different configurations (a total of 4). The first two use in-memory trace exporter, and the last two use Cloud Trace backend.

ehsannas: Regarding `the default behaviour is being changed to "use custom otel or fall back to Global otel", do we still need that 4 test classes?`

We still need all these tests.

`ITE2ETracingTestGlobalOtel` and `ITTracingTestGlobalOtel` ensure that we collect traces when a global OTEL is registered, _but not passed to Firestore_.

`ITE2ETracingTestNonGlobalOtel` and `ITTracingTestNonGlobalOtel` ensure that we collect traces using the OTEL instance that has been passed to Firestore.

milaGGL (approved):
LGTM

PR #1896, opened 2024-10-21
fix: make StreamableQuery#internalStream package-private to avoid documentation and prevent clirr errors

MarkDuckworth:


google-cloud-firestore/src/main/java/com/google/cloud/firestore/StreamableQuery.java
tom-andersen, line 1: Could we make the entire class package private?

MarkDuckworth: It's possible that customers have declared variables of this type. `StreamableQuery query = firestore.collection(...).where(...)`. I can deprecate use of this class and remove it from the API surface in the next major release.

MarkDuckworth: Instead we decided to apply the `@InternalApi` attribute to the StreamableQuery base class.

tom-andersen (approved):
This aligns with the direction I want to move.

Note: I would have gone further, but in the exact same direction, so LGTM and let's get this merged.

PR #1935, opened 2024-11-19
feat: introduce `java.time` variables and methods

diegomarquezp:
This PR introduces `java.time` alternatives to existing `org.threeten.bp.*` methods, as well as switching internal variables (if any) to `java.time`

The main constraint is to keep the changes backwards compatible, so for each existing threeten method "`method1(org.threeten.bp.Duration)`" we will add an alternative with a _Duration_ (or _Timestamp_ when applicable) suffix: "`method1Duration(java.time.Duration)`".

For most cases, the implementation will be held in the `java.time` method and the old threeten method will just delegate the call to it. However, for the case of abstract classes, the implementation will be kept in the threeten method to avoid breaking changes (i.e. users that already overloaded the method in their user code).

### DIFF
--- a/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
+++ b/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
@@ -16,11 +16,13 @@
 
 package com.google.cloud.firestore;
 
+import static com.google.api.gax.util.TimeConversionUtils.toThreetenDuration;
 import static com.google.cloud.firestore.telemetry.TraceUtil.*;
 
 import com.google.api.core.ApiClock;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.NanoClock;
+import com.google.api.core.ObsoleteApi;
 import com.google.api.core.SettableApiFuture;
 import com.google.api.gax.rpc.ApiStreamObserver;
 import com.google.api.gax.rpc.BidiStreamObserver;
@@ -52,7 +54,6 @@ import java.util.Map;
 import java.util.Random;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import org.threeten.bp.Duration;
 
 /**
  * Main implementation of the Firestore client. This is the entry point for all Firestore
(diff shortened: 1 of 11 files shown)


google-cloud-firestore/src/main/java/com/google/cloud/firestore/telemetry/EnabledTraceUtil.java
```diff
@@ -94,12 +96,16 @@ public ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder> getChannelConfi
...
+    return durationStringDuration(toJavaTimeDuration(duration));
+  }
+
   // Returns a JSON String representation of the given duration. The JSON representation for a
   // Duration is a String that
   // ends in `s` to indicate seconds and is preceded by the number of seconds, with nanoseconds
   // expressed as fractional
   // seconds.
-  String durationString(org.threeten.bp.Duration duration) {
+  String durationStringDuration(java.time.Duration duration) {
```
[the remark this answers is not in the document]
diegomarquezp: Good catch. Overload removed

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
```diff
@@ -500,9 +501,16 @@ public FirestoreRpc getClient() {
...
   @Override
-  public Duration getTotalRequestTimeout() {
-    return firestoreOptions.getRetrySettings().getTotalTimeout();
+  public org.threeten.bp.Duration getTotalRequestTimeout() {
+    return toThreetenDuration(getTotalRequestTimeoutDuration());
+  }
+
+  @Override
+  public java.time.Duration getTotalRequestTimeoutDuration() {
+    return firestoreOptions.getRetrySettings().getTotalTimeoutDuration();
```
lqiu96, line 513: nit: Would be nice to get javadocs from firestore for this

lqiu96 (approved):
LGTM. Firestore team for final review

ehsannas (approved):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java
diegomarquezp: @ehsannas for visibility - the superclass does not have javadocs on this method

PR #2021, opened 2025-02-18
FIX: Use lazy encoding in UTF-8 encoded string comparison

milaGGL:
The previous [fix](https://github.com/googleapis/java-firestore/pull/1967/files) created a performance issue due to expensive UTF-8 encoding. Update `compareUtf8Strings` to use lazy encoding instead.


google-cloud-firestore/src/main/java/com/google/cloud/firestore/Order.java
```diff
@@ -136,9 +136,31 @@ public int compare(@Nonnull Value left, @Nonnull Value right) {
...
+      int rightCodePoint = right.codePointAt(i);
+
+      if (leftCodePoint != rightCodePoint) {
+        if (leftCodePoint < 128 && rightCodePoint < 128) {
+          // ASCII comparison
+          return Integer.compare(leftCodePoint, rightCodePoint);
+        } else {
+          // UTF-8 encoded byte comparison, substring 2 indexes to cover surrogate pairs
+          ByteString leftBytes =
+              ByteString.copyFromUtf8(left.substring(i, Math.min(i + 2, left.length())));
```
dconeybe, line 151: Should this be `left.length() - i`?

dconeybe (dismissed):

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Order.java
ehsannas: Nit: @milaGGL I would improve the following comment to be more descriptive (along the lines of what you said above):

changing
```java
// UTF-8 encoded byte comparison, substring 2 indexes to cover surrogate pairs
```
to
```java
// We have identified a code point difference, so we don't need to
// encode/compare _all_ the remainder of the strings. We only need to 
// compare up to 2 more indexes to cover potential surrogate pairs.
```

google-cloud-firestore/src/main/java/com/google/cloud/firestore/Order.java
```diff
@@ -136,9 +136,39 @@ public int compare(@Nonnull Value left, @Nonnull Value right) {
...
+
+  private static String getUtf8SafeBytes(String str, int index) {
+    int firstCodePoint = str.codePointAt(index);
+    if (firstCodePoint > 0xffff) {
+      // It's a surrogate pair, return the whole pair
+      return str.substring(index, index + 2);
+    } else {
+      // It's a single code point, return it
+      return str.substring(index, index + 1);
+    }
```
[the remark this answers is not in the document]
milaGGL: yes, we can do this for java and android. But, js doesn't have equivalent function, so I will keep web and node unchanged.

ehsannas (approved):

PR #2055, opened 2025-03-21
fix: elide string processing of 'fine' level log messages unless they will actually be logged

### LINKED ISSUE #2048: FINE logging of toString taking up a significant amount of CPU
For instance this line https://github.com/googleapis/java-firestore/blob/main/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Watch.java#L477.

It does a toString on the entire snapshot and then discards it. The common pattern when doing expensive logging like this is to check if the log level is enabled before doing the heavy toString

dconeybe:
Fixes #2048


tom-andersen (approved):
Subtle but nice!

PR #2157, opened 2025-06-23
test: Test large documents.

ehsannas:


ehsannas:
This was ported from https://github.com/firebase/firebase-android-sdk/pull/7040

PR #2369, opened 2026-04-17
ci: update hermetic generation workflow to use the latest image

suztomo:
We moved the definition of the workflow files from sdk-platform-java repository to the sdk-platform-java folder in the google-cloud-java repository.

This uses the "latest" Docker tag in .github/workflows/hermetic_library_generation.yaml to confirm the behavior of the today's hermetic build container image update.

b/503444342

### DIFF
--- a/.github/workflows/hermetic_library_generation.yaml
+++ b/.github/workflows/hermetic_library_generation.yaml
@@ -40,6 +40,7 @@ jobs:
     - uses: googleapis/google-cloud-java/sdk-platform-java/.github/scripts@v1.85.0
       if: env.SHOULD_RUN == 'true'
       with:
+        image_tag: latest
         base_ref: ${{ github.base_ref }}
         head_ref: ${{ github.head_ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/hermetic_library_generation.yaml
```diff
@@ -40,6 +40,7 @@ jobs:
     - uses: googleapis/google-cloud-java/sdk-platform-java/.github/scripts@v1.85.0
       if: env.SHOULD_RUN == 'true'
       with:
+        image_tag: latest
```
blakeli0, line 43: I don't think we want to point the image to latest. It was designed to use a release tag to prevent accidental changes. 
I assume this is only for testing and will be removed before merging?

blakeli0 (changes requested):

.github/workflows/hermetic_library_generation.yaml
suztomo: Ack.

suztomo:
Closing for https://buganizer.corp.google.com/issues/503444342#comment5

